PR #227, opened 2022-09-19
[MANOPD-77802] - Check default thirdparties sha despite of source

Imadzuma:
### Description
We've already had default sha for thirdparties, but we specially delete them. And then we skip sha check, if it's not defined in cluster.

Fixes # (issue)
* If default sha is not defined, for check caches thirdparty will be loaded to one of control-plane node and expexted sha will be got from this file.

### How to apply
* [Install task/s](documentation/Installation.md#installation-tasks-description)
* [Check Paas](documentation/Kubecheck.md#paas-procedure)

### Checklist
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/kubemarine/procedures/check_paas.py
+++ b/kubemarine/procedures/check_paas.py
@@ -299,6 +299,7 @@ def thirdparties_hashes(cluster):
     """
     with TestCase(cluster.context['testsuite'], '212', "Thirdparties", "Hashes") as tc:
         successful = []
+        warnings = []
         broken = []
 
         #Create tmp dir for loading thirdparty without default sha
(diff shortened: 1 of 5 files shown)


kubemarine/thirdparties.py
```diff
@@ -79,12 +79,6 @@ def enrich_inventory_apply_defaults(inventory, cluster):
                                     'Expected any of %s, but \'%s\' found.'
                                     % (destination, all_nodes_names, node_name))
 
-        # if source is re-defined by user, but "sha1" is not provided,
-        # then remove default "sha1", because it may be wrong
-        raw_config = raw_inventory.get('services', {}).get('thirdparties', {}).get(destination, {})
-        if 'source' in raw_config and 'sha1' not in raw_config and 'sha1' in config:
-            del config['sha1']
```
theboringstuff, line 86: Is it supported for our default thirdparties (e.g. calicoctl, crictl) that the user can define their own versions? Like, we have `crictl-1.2.3` in our defaults, but user specifies `crictl-1.2.4`. If this is valid, then I think our default SHAs can lead to a situation where user specifies `crictl-1.2.4` (+ we insert default SHA for `crictl-1.2.3`), but at the same time user already has `crictl-1.2.3` on nodes. In this case, we will compare default SHA to the node SHA and they will be equal (1.2.3), so we will successfully say that 1.2.4 is installed.

Imadzuma: As we discussed, kubemarine should download thirdparty from source and get sha from it.
This way was implemented.

kubemarine/procedures/check_paas.py
```diff
@@ -309,9 +313,45 @@ def thirdparties_hashes(cluster):
...
+                                expected_sha = result.stdout.strip()
+                                cluster.log.verbose("Expected sha was got")
+                    # Remove temporary dir in any case
+                    cluster.log.verbose(f"Remove temporary dir {random_dir}...")
+                    first_control_plane.sudo(final_commands, hide=True, warn=True)
+
+                if expected_sha is None:
+                    cluster.log.verbose(f"Can`t get expected sha for {path}, skip it")
+                    #Skip checking sha if something went wrong or this sha can't be loaded
+                    continue
```
ilia1243, line 351: For default thirdparties, we might additionally compare computed expected_sha with the recommended default sha and raise warning if they are not equal.

ilia1243: Sorry, probably I said not accurate enough. I meant sha of default thirdparties which we recommend for each kubernetes version. For example [kubeadm](https://github.com/Netcracker/KubeMarine/blob/0.7.1/kubemarine/resources/configurations/defaults.yaml#L82). We can compare that sha with calculated expected_sha on first_control_plane.

In other words, let's assume user specified custom source for `kubeadm` and not specified custom sha. In this case config['sha'] will be removed. We expect that it is different source but the same thirdparty version, because this version is QA tested. Other versions are allowed but we can warn if they are used. @koryaga do we need to do that?

Imadzuma: @ilia1243 recommended 3dparty version was calculated additionally in paas_check. Please check

kubemarine/procedures/check_paas.py
```diff
@@ -308,10 +313,54 @@ def thirdparties_hashes(cluster):
...
+                # Remove temporary dir in any case
+                cluster.log.verbose(f"Remove temporary dir {random_dir}...")
+                first_control_plane.sudo(final_commands, hide=True, warn=True)
 
+            recommended_sha = thirdparties.get_thirdparty_recommended_sha(path, cluster)
+            if recommended_sha is not None and recommended_sha != expected_sha:
+                warnings.append(f"{path} source contains not recommended thirdparty version for used kubernetes version")
+
+            if config.get("sha1", expected_sha) != expected_sha:
+                warnings.append("Given sha is not equal with actual sha from source for %s" % path)
```
ilia1243, line 355: I think this can be considered as broken rather than warning.

kubemarine/procedures/check_paas.py
```diff
@@ -308,10 +313,54 @@ def thirdparties_hashes(cluster):
...
+                            broken.append(f'failed to get sha for temporary file {random_path} on {host.host}: {result.stderr}')
+                            cluster.log.verbose(f'failed to get sha for temporary file {random_path} on {host.host}: {result.stderr}')
+                        else:
+                            expected_sha = result.stdout.strip()
+                            cluster.log.verbose(f"Expected sha was got for {path}")
+                # Remove temporary dir in any case
+                cluster.log.verbose(f"Remove temporary dir {random_dir}...")
+                first_control_plane.sudo(final_commands, hide=True, warn=True)
 
+            recommended_sha = thirdparties.get_thirdparty_recommended_sha(path, cluster)
```
ilia1243, line 350: Now we have two independent algorithms to calculate recommended sha. In defaults using jinja templates and in python code. So why not keep the single algorithm then? Just remove sha sections from defaults.yaml and enrich them in thirdparties.enrich_inventory_apply_defaults

kubemarine/thirdparties.py
```diff
@@ -38,6 +38,29 @@ def enrich_inventory_apply_upgrade_defaults(inventory, cluster):
...
+    # Get kubeadm version
+    kubernetes_version = cluster.inventory['services']['kubeadm']['kubernetesVersion']
+    effective_kubernetes_version = ".".join(kubernetes_version.split('.')[0:2])
+
+    # Get software versions map
+    software_name = cluster.globals['thirdparties'].get(destination, {}).get('software_name', None)
+    software_versions = cluster.globals['compatibility_map']['software'].get(software_name, {})
+
+    # Return sha1 related to used kubernetes version
+    recommended_sha = software_versions[kubernetes_version].get('sha1', None) \
```
ilia1243, line 52: Can this throw `KeyError: kubernetes_version` for not default 3rd-parties?

kubemarine/resources/configurations/globals.yaml
```diff
@@ -102,6 +102,18 @@ workaround:
   retries: 10
   delay_period: 5
 
+thirdparties:
```
ilia1243, line 105: If new default 3rd-party is added, but this section is forgotten to update, check_paas will consider 3rd-party as not default and won't check recommended versions. It will not even fail in this case, right?
On the other hand, we are starting to move the list of default 3rd-parties from defaults to globals and can complete this rework.
* Move the whole section without `sha` from default to globals.
* Add new enrichment procedure before defaults.compile_inventory which will
    * Copy default 3rd-party from global to inventory, if it is not defined by user.
    * Added recommended sha using thirdparties.get_thirdparty_recommended_sha if source is not defined by user.
* check_paas will use the same thirdparties.get_thirdparty_recommended_sha

kubemarine/thirdparties.py
```diff
@@ -38,6 +38,29 @@ def enrich_inventory_apply_upgrade_defaults(inventory, cluster):
     return inventory
 
 
+def get_thirdparty_recommended_sha(destination, cluster):
```
ilia1243, line 41: I think we can add unit tests for this method
* default 3rd-party which requires N.N.N k8s version
* default 3rd-party which requires N.N k8s version
* non-default 3rd-part

kubemarine/resources/configurations/globals.yaml
Imadzuma: If we forget to add new 3dparty there, kubemarine can't calculate recommended sha for it and will work with it like with custom 3dparty: sha checks will be only if user specify it.
About moving thirdparties from defaults.yaml to globals.yaml: we can do it, of course, but I think, it breaks the appointment of those files:
* defaults.yaml contains default values in cluster.yaml format and those values can be overridden directly.
* globals.yaml contains values constants that are not available for user customization. 
Thirdparties section is more suitable for defaults.yaml for this appointment. 
I think, this is out of scope for this bug and we need additional discussion for it

ilia1243: There are few precedents that globals have some defaults that can be redefined. I found logging and `prepull_group_size`.
Ok, if we want to move the discussion to the separate task and PR, we can skip checking of recommended version for now.

kubemarine/procedures/check_paas.py
Imadzuma: Yes, because we compare sha with given from user. Fixed

kubemarine/procedures/check_paas.py
Imadzuma: Default sha were calculated from function. sha fields from defaults.yaml were deleted

kubemarine/thirdparties.py
Imadzuma: No, because it's ternary operator and this line is used if kubernetes_version key exists in software_versions dict.
See the next line

kubemarine/thirdparties.py
Imadzuma: Tests were added for calculating recommended sha and sha in inventory

ilia1243 (approved):

PR #233, opened 2022-09-28
MANOPD-76951 Fix "remote_node" procedure for miniHA

Corey-666:
### Description
Job fails if remove **shutdowned** node using maintenance procedure for **miniHA** scheme. When preserving of the information about the procedure the actual number of nodes is not taken into account.

### Solution
Additionally, use the **get_final_nodes()** function to get the actual list of nodes.

### How to apply
none

### Test Cases

**TestCase 1**

Test Configuration:
miniHA cluster of 3 master nodes

Steps:
1. Shutdown master node
2. Remove node from cluster using maintenance procedure
3. The node removal procedure will be fully executed, but the job will fail at the final stage of generating logs about the performed procedure. 

Results:

| Before | After |
| ------ | ------ |
| Job successfully finished | Job failure with python traceback error |

### Checklist
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/kubemarine/core/utils.py
+++ b/kubemarine/core/utils.py
@@ -342,7 +342,7 @@ class ClusterStorage:
                 for i, file in enumerate(files):
                     if i >= not_pack_file and i < delete_old:
                         if 'tar.gz' not in file:
-                            control_plane.get_final_nodes().sudo(f'tar -czvf {self.dir_path + file + ".tar.gz"} {self.dir_path + file} &&'
+                            control_plane.sudo(f'tar -czvf {self.dir_path + file + ".tar.gz"} {self.dir_path + file} &&'
                                        f'sudo rm -r {self.dir_path + file}')
                     elif i >= delete_old:
                         control_plane.sudo(f'rm -rf {self.dir_path + file}')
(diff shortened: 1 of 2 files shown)


kubemarine/core/utils.py
```diff
@@ -342,7 +342,7 @@ def rotation_file(self):
                 for i, file in enumerate(files):
                     if i >= not_pack_file and i < delete_old:
                         if 'tar.gz' not in file:
-                            control_plane.sudo(f'tar -czvf {self.dir_path + file + ".tar.gz"} {self.dir_path + file} &&'
+                            control_plane.get_final_nodes().sudo(f'tar -czvf {self.dir_path + file + ".tar.gz"} {self.dir_path + file} &&'
```
ilia1243, line 345: Here control_plane is a single node which is surely not a node to be removed. So the change is unnecessary.

kubemarine/core/utils.py
```diff
@@ -386,7 +386,7 @@ def collect_info_all_control_plane(self):
         This method is used to transfer backup logs from the main control-plane to the new control-plane.
         """
 
-        node = self.cluster.nodes['control-plane'].get_initial_nodes().get_first_member(provide_node_configs=True)
+        node = self.cluster.nodes['control-plane'].get_final_nodes().get_initial_nodes().get_first_member(provide_node_configs=True)
```
ilia1243, line 389: This method is invoked only during add_node procedure, so change is not necessary.

kubemarine/core/utils.py
Corey-666: Fixed

kubemarine/core/utils.py
Corey-666: fixed too

ilia1243 (approved):

PR #243, opened 2022-10-04
Fix calico paas check

n549:
### Description
In 0.7.1 check_paas procedure starts failing on the 224 check (Calico configfuration check).

Fixes # (issue)

### Solution
New approach to calico envs definition is introduced in v0.7.1. According to that an env can look like
```
      NODENAME:
          fieldRef:
            fieldPath: spec.nodeName
```
which doesn't fit to the condition https://github.com/Netcracker/KubeMarine/blob/dcc7040664fad54cbfcf2e8dacbb8abe996f3bf1/kubemarine/procedures/check_paas.py#L1058
```
                if not str(cluster.inventory["plugins"]["calico"]["env"].get(env["name"])) == env["value"]:
```

### Test Cases
Run check_paas procedure. The check 224 should run with "valid" result.

### Checklist
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/kubemarine/procedures/check_paas.py
+++ b/kubemarine/procedures/check_paas.py
@@ -1054,7 +1054,7 @@ def calico_config_check(cluster):
         result = result.get_simple_out()
         result = yaml.safe_load(result)
         for env in result["spec"]["template"]["spec"]["containers"][0]["env"]:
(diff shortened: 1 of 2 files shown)


andrewluckyguy (approved):
Retested by QA

PR #252, opened 2022-10-12
Ubuntu LTS 22.04.1 support

dmyar21:
### Description
* Need to support the new version of ubuntu 22.04.1

### Solution
* Added support for ubuntu 22.04.1 in global.yaml
* Edit Installation.md
* Fix template haproxy.cfg and procedure install haproxy on the cluster
* Fix procedure add_node.py
* Support new version Docker-CE (20.10)

### How to apply
* Update cri Docker
     Run procedure install using a --tasks="prepare.cri.install"
### Test Cases

**Install case**

Test Configuration:

- Hardware: All-in-one
- OS: Ubuntu 22.04.1
- Inventory: common

Steps:

1. Install

Results:

| Before | After |
| ----- | ----- | 
| Not supported OS  | Supported OS  |

### Checklist
- [ ] Integration CI passed

#### Unit tests


kubemarine/procedures/add_node.py
```diff
@@ -101,10 +101,13 @@ def cache_installed_packages(cluster):
 
     # avoid caching when unchanged nodes are not equal to GLOBAL os
     global_os = system.get_os_family(cluster)
-    for node in cluster.nodes['all'].get_unchanged_nodes().get_ordered_members_list(provide_node_configs=True):
-        if cluster.context["nodes"][node['connect_to']]["os"]['family'] != global_os:
+    for nodes in cluster.nodes['all'].get_first_member().get_ordered_members_list(provide_node_configs=True):
+        version_os = cluster.context["nodes"][nodes['connect_to']]["os"]['version']
+    for node in cluster.nodes['all'].get_changed_nodes().get_ordered_members_list(provide_node_configs=True):
+        if cluster.context["nodes"][node['connect_to']]["os"]['family'] != global_os \
+                or cluster.context["nodes"][node['connect_to']]["os"]['version'] != version_os:
```
ilia1243, line 108: Previously only unchanged nodes were compared. Why comparing changes nodes now?

ilia1243, line 108: The idea is good, but

1. It looks strange to take global_os and version_os from different sources but use the similar way.
2. It does not make sense to compare changed_nodes (new_nodes actually) and global_os, as global_os is calculated based on new nodes only. So the != comparison result is always false.
3. The following case is not covered: initial nodes have OS2 and OS1. New node has OS2. version_os fill be detected by first node which is OS2. The != comparison with changed (new) nodes will be false. Packages are cached but isn't it  a problem that there are finally cluster with different OS: OS2, OS1, OS2?

kubemarine/procedures/add_node.py
```diff
@@ -101,10 +101,13 @@ def cache_installed_packages(cluster):
...
             cluster.log.debug(f"New node has different OS ({global_os}) "
-                              f"than some other nodes ({cluster.context['nodes'][node['connect_to']]['os']['family']}), "
+                              f"than some other nodes ({cluster.context['nodes'][node['connect_to']]['os']['version']}), "
                               "packages will not be cached.")
```
ilia1243, line 111: Previously the message said about difference of os family. Not it compares os family and version that does not make sense.

ilia1243 (changes requested):

kubemarine/procedures/add_node.py
dmyar21: Fix code

kubemarine/procedures/add_node.py
dmyar21: add new information

kubemarine/procedures/add_node.py
```diff
@@ -99,15 +99,35 @@ def cache_installed_packages(cluster):
...
+            cluster.log.debug(
+                f"New node has different OS Version ({cluster.context['nodes'][node['connect_to']]['os']['family']},"
+                f"{cluster.context['nodes'][node['connect_to']]['os']['version']}) "
+                f"than some other nodes, "
+                "packages will not be cached.")
+            return
+        elif cluster.context["nodes"][node['connect_to']]["os"]['family'] in family_os \
+             and cluster.context["nodes"][node['connect_to']]["os"]['version'] in version_os:
+            for old_version_os in version_os:
+                if cluster.context["nodes"][node['connect_to']]["os"]['version'] != old_version_os:
```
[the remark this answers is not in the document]
ilia1243: More simple algorithm: compare os_family and os_version of each nodes in nodes['all'] with each other. No need to use get_changed_nodes or get_unchanged_nodes. Looks like it solves all problems.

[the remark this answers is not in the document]
ilia1243: Or even simpler, use set of tuples of family and version and compare if len() > 1: return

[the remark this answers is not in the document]
dmyar21: use set of tuples of family and version and compare

kubemarine/resources/configurations/globals.yaml
```diff
@@ -108,43 +108,43 @@ compatibility_map:
       v1.21.2:
         version_rhel: 19.03*
         version_rhel8: 19.03*
-        version_debian: 5:19.03.*
+        version_debian: 5:20.10.*
```
ilia1243, line 111: If we upgrade recommendend versions for existing clusters, we need to create patch to migrate the clusters

kubemarine/haproxy.py
```diff
@@ -72,7 +72,7 @@ def install(group):
     with RemoteExecutor(group.cluster) as exe:
         for node in group.get_ordered_members_list(provide_node_configs=True):
             package_associations = group.cluster.get_associations_for_node(node['connect_to'])['haproxy']
-            group.sudo("%s -v" % package_associations['executable_name'], warn=True)
+            group.sudo("bash %s -v" % package_associations['executable_name'], warn=True)
```
ilia1243, line 75: We need to find root cause. I do not know what happens on ubuntu 22, but on 20 it stops working
```
ubuntu@k8s-balancer-left:~$ /usr/sbin/haproxy -v
HA-Proxy version 2.0.29-0ubuntu1 2022/08/26 - https://haproxy.org/
ubuntu@k8s-balancer-left:~$ bash /usr/sbin/haproxy -v
/usr/sbin/haproxy: /usr/sbin/haproxy: cannot execute binary file 
```

kubemarine/resources/configurations/globals.yaml
dmyar21: New versions of packages will be used for new nodes, I don’t really see the point in creating a patch

ilia1243: It is possible to run install --tasks prepare.cri or other task which in particular installs packages on existing clusters. There can be other reasons to run the tasks. In general patches are developed to make cluster consistent with the new version of kubemarine. So I suggest to emphasize that new version of kubemarine requires new packages, using How to Apply, and either automatic patch or manual steps.

dmyar21: Add how to apply

ilia1243 (approved):

Issue #245, opened 2022-10-07
root user in the Dockerfile


alekslebedev:
Hi, team!
Using a **root** user in the [Dockerfile](https://github.com/Netcracker/KubeMarine/blob/9369b0edf02d8b25a41cb396decbef4094c44f00/Dockerfile#L8) is not a best practice. Is this really required ?

koryaga:
#256

koryaga:
resolved in 0.9.0

PR #256, opened 2022-10-17
Update Dockerfile

nikhil1697:
### Description
Using a root user in the Dockerfile is not a best practice, removing USER as root in Dockerfile did not make any noticeable changes so removed it from the dockerfile 

### Solution
Removed USER as root


koryaga:
#245

koryaga:
@nikhil1697  pls update PR description accroding to template.

koryaga (approved):

PR #268, opened 2022-10-27
[MANOPD-80822] change default typha settings

n549:
### Description
For medium and large clusters it's recommended to use `typha` in `calico` deployments to decrease number of calls to kubernetes api-server.
This PR is to change default `calico` settings.

Fixes # (issue)
MANOPD-80822

### Solution
Default `calico` settings are changed. `typha` is enabled by default for clusters with more than 3 nodes and at least 2 replicas of `typha` are created for clusters with 4 nodes and more. With every 50 nodes number of `typha` replicas is incremented.

2 additional jinja filters were created: `is_true` and `is_false`. They are added to the calico templates in the conditions which test `calico.typha.enabled` value. It's necessary because yaml parser considers `'true'` value as a string, not as a boolean variable.

### Test Cases
1. Deploy `calico` with default settings to the cluster with <= 3 nodes. Check that deployment finishes successfully, no `typha` installed, pods at different nodes can see each other.
2. Deploy `calico` with default settings to the cluster with more than 3 nodes. Check that deployment finishes successfully, `typha` pods are running and ready, pods at different nodes can see each other.

### Checklist
- [ ] Integration CI passed

### DIFF
--- a/kubemarine/jinja.py
+++ b/kubemarine/jinja.py
@@ -28,8 +28,8 @@ def new(log, root=None):
     env.filters['majorversion'] = lambda version: precompile(log, version, root).split('.')[0]
     # we need these filters because rendered cluster.yaml can contain variables like 
     # enable: 'true'
-    env.filters['is_true'] = lambda v: True if (v in ['true','True','TRUE'] or v == True) else False
-    env.filters['is_false'] = lambda v: True if (v in ['false','False','FALSE'] or v == False) else False
+    env.filters['is_true'] = lambda v: v in ['true', 'True', 'TRUE', True]
+    env.filters['is_false'] = lambda v: v in ['false', 'False', 'FALSE', False]
     return env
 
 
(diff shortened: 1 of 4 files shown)


kubemarine/jinja.py
```diff
@@ -26,7 +26,10 @@ def new(log, root=None):
     env.filters['isipv4'] = lambda ip: ":" not in precompile(log, ip, root)
     env.filters['minorversion'] = lambda version: ".".join(precompile(log, version, root).split('.')[0:2])
     env.filters['majorversion'] = lambda version: precompile(log, version, root).split('.')[0]
-
+    # we need these filters because rendered cluster.yaml can contain variables like 
+    # enable: 'true'
+    env.filters['is_true'] = lambda v: True if (v in ['true','True','TRUE'] or v == True) else False
+    env.filters['is_false'] = lambda v: True if (v in ['false','False','FALSE'] or v == False) else False
```
ilia1243, line 32: Minor point, it can be simpler
```suggestion
    env.filters['is_true'] = lambda v: v in ['true', 'True', 'TRUE', True]
    env.filters['is_false'] = lambda v: v in ['false', 'False', 'FALSE', False]
```

ilia1243 (approved):

kubemarine/jinja.py
n549: thanks, updated

alexarefev (approved):

PR #278, opened 2022-11-09
MANOPD-81038 MANOPD-78302 Refactoring and fixes of OS/package versions

### LINKED ISSUE #262: Unclear flow to work with nodes OS
Currently we have complex algorithm to work with nodes OS which is hard to understand:
* cluster.context["os"] is assigned differently for add_node and other procedures. For add_node it is calculated based only on new nodes. 
* NodeGroup.get_nodes_os() performs validation and might throws exception based on `suppress_exceptions` parameter. It is hard to understand when the getter will throw exception and when it will not.
* NodeGroup.get_nodes_os() accepts `force_all_nodes` parameter which makes it even more complicated.

I suggest to:
* get rid of cluster.context["os"]. Instead, we can always calculate nodes OS using NodeGroup.get_nodes_os() for necessary nodes group (only new nodes for add_node or for all nodes in other procedures).
* get rid of both parameters `suppress_exceptions` and `force_all_nodes` in NodeGroup.get_nodes_os(). It is necessary to have single place where validation should happen and error is thrown. Instead of `node_group.get_nodes_os(force_all_nodes=False)` it is possible to use more clear `node_group.get_new_nodes_or_self().get_nodes_os()`

LOE: 3-5 md

ilia1243:
### Description
1. `cluster_finalized.yaml` no longer caches versions of custom packages from `services.packages.install` section after #168
1. Changes in enrichment procedure in #181 related to `cache_versions` property are doubtful and should be reverted/reworked.
    * If the user customizes an association, its caching is skipped during add_node.
      Instead, turning the caching off should be done intentionally via the same `cache_versions=false` property for each necessary association.
    * If the user uses `cache_versions=false` for all associations, we do not use our recommended package versions.
      We should do use recommended versions.
      The user will still be able to not follow our recommendations by explicitly redefining of `package_name` section of each necessary association.
1. Detection and work with OS versions are weird and should be reworked.

Fixes #262

### Solution
* Enrichment procedure is changed as per the p. 2 of the Description.
* `cluster_finalized.yaml` now caches package versions independently on `cache_versions` property.
  If any package is installed with different versions on different nodes, warning is printed and the package is taken as-is from the inventory.
  If multiple OS families or versions detected on nodes, packages detection is skipped.
* Major refactoring of OS and package versions detection, enrichment and usage.
    * Enrichment of packages now happens independently on whether it is `add_node` or any other procedure.
      Customized packages are moved to OS family specific section and later used from there.
    * It is **no longer possible** to customize associations in global section for clusters in multiple OS.
      The customization should be moved to the corresponding OS specific section before OS migration.
    * `cluster.context['os']` variable is removed. Now cluster global OS is detected using `cluster.get_os_family()`.
      It does not depend on the procedure with the only exception that global OS family is calculated based on final nodes only.
    * `NodeGroup.get_nodes_os()` no longer accepts any flag.
      It returns OS family of really those group on which it is called and does not throw exceptions for unsupported or multiple OS families.
* Fixed unnecessary audit installation if balancer is added.
* Resolving of `cluster_finalized.yaml` is rewritten to enrichment-like style.
* Other fixes of `cluster_finalized.yaml` resolving if multiple OS families or versions detected on nodes or if some nodes are not sudoers.
* Fixes in `check_paas` procedure for clusters on multiple OS.
* Other refactoring and fixes.

### Test Cases

**TestCase 1**

Test Configuration:

- Hardware: Any
- OS: All nodes have the same OS
- Inventory: Any with non-empty `services.packages.install` section. Some of packages have equal version on all nodes and some of them do not.

Steps:

1. Run any procedure

Results:

| Before | After |
| ------ | ------ |
| `cluster_finalized.yaml` does not contain versions for `services.packages.install` section | Packages that have equal version on all nodes are cached. For other packages there are warnings in logs. |

### Regression Tests
* `upgrade` and `migrate_cri` procedures with custom `packages` sections.
* Audit installation on cluster with debian nodes.
* OS migration
* `add_node` when initial nodes having different associations packages versions and if `cache_versions` option is specified.

### Checklist
- [x] I have made corresponding changes to the documentation.
- [x] Integration CI passed
- [x] Unit tests. If Yes list of new/changed tests with brief description.

#### Unit tests
* test_cluster.py: added tests for cluster.get_os_family()
* test_audit.py, test_haproxy.py, test_keepalived.py: fixes according to new enrichment approach
* test_migrate_cri.py, test_upgrade.py: added tests for packages enrichment and finalization
* test_packages.py: added many tests for packages enrichment, detection and caching.

### DIFF
--- a/kubemarine/packages.py
+++ b/kubemarine/packages.py
@@ -12,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from copy import deepcopy
 from typing import List, Dict
 
-from kubemarine import yum, apt, system
+from kubemarine import yum, apt
 from kubemarine.core.cluster import KubernetesCluster
 from kubemarine.core.executor import RemoteExecutor
 from kubemarine.core.group import NodeGroup, NodeGroupResult
@@ -23,26 +22,26 @@ from kubemarine.core.yaml_merger import default_merger
 
 
 def enrich_inventory_associations(inventory, cluster: KubernetesCluster):
-    os_family = cluster.get_os_family_from_new_nodes_or_final()
-    # todo what if new and existing nodes have different OS family but new nodes have the same os family?
-    if os_family in ('unknown', 'unsupported', 'multiple'):
-        cluster.log.debug("Skip enrichment of associations as the nodes do not have single OS family")
-        # Skip enrichment. Some features might not work.
-        return inventory
-
     associations: dict = inventory['services']['packages']['associations']
-
-    # copy associations for OS family to one level higher
-    os_specific_associations = deepcopy(associations[os_family])
+    os_propagated_associations = {}
 
     # move associations for OS families as-is
     for association_name in get_associations_os_family_keys():
-        os_specific_associations[association_name] = associations.pop(association_name)
+        os_propagated_associations[association_name] = associations.pop(association_name)
 
-    # merge remained explicitly defined package properties with priority
-    default_merger.merge(os_specific_associations, associations)
+    inventory['services']['packages']['associations'] = os_propagated_associations
 
-    inventory['services']['packages']['associations'] = os_specific_associations
+    # Check remained associations section if they are customized at global level.
+    if associations:
+        os_family = cluster.get_os_family()
+        if os_family == 'multiple':
+            raise Exception(
+                "It is not supported to customize services.packages.associations section "
+                "if nodes have different OS families. "
+                "Please move the section to corresponding services.packages.associations.<os_family> section.")
+        elif os_family not in ('unknown', 'unsupported'):
+            # move remained associations properties to the specific OS family section and merge with priority
+            default_merger.merge(os_propagated_associations[os_family], associations)
 
     return inventory
 
@@ -52,7 +51,7 @@ def cache_package_versions(cluster: KubernetesCluster, inventory: dict, ensured_
     different_os = list(set(os_ids.values()))
     if len(different_os) > 1:
         cluster.log.debug(f"Final nodes have different OS families or versions, packages will not be cached. "
-                       f"List of (OS family, version): {different_os}")
+                          f"List of (OS family, version): {different_os}")
         return inventory
 
     os_family = different_os[0][0]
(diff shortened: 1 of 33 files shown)


kubemarine/packages.py
```diff
@@ -139,92 +266,45 @@ def detect_installed_package_version(group: NodeGroup, package: str, warn=True)
...
-
-    detected_packages = detect_installed_packages_versions(group, packages_list)
-    grouped_packages: Dict[str, Dict[str, List]] = {}
-    for queried_package, detected_packages_results in detected_packages.items():
-        detected_grouped_packages = {}
-        for host, packages in detected_packages_results.items():
-            if '\n' in packages:
-                # this is the test, when package name contains multiple names,
-                # e.g. docker-ce and docker-cli for "docker-ce-*" query
-                packages = packages.split('\n')
```
ilia1243, line 215: We no longer query like `docker-ce-*`.

kubemarine/packages.py
```diff
@@ -12,64 +12,191 @@
...
+    return associated_packages
+
+
+def _get_packages_for_associations_to_detect(cluster: KubernetesCluster, inventory: dict, association_name: str,
+                                             ensured_association_only: bool) -> list:
+    packages_list = _get_package_names_for_association(cluster, inventory, association_name)
+    if not packages_list:
+        return []
+
+    global_cache_versions = inventory['services']['packages']['cache_versions']
```
ilia1243, line 104: I suggest to rename the option to `ensure_same_versions` as it now works only for `add_node` and can be thought of as we guarantee same versions on nodes. What do you think?

kubemarine/packages.py
```diff
@@ -12,64 +12,196 @@
...
+    for association_name in get_associations_os_family_keys():
+        os_propagated_associations[association_name] = associations.pop(association_name)
+
+    inventory['services']['packages']['associations'] = os_propagated_associations
+
+    # Check remained associations section if they are customized at global level.
+    if associations:
+        os_family = cluster.get_os_family()
+        if os_family == 'multiple':
+            raise Exception(ERROR_GLOBAL_ASSOCIATIONS_REDEFINED_MULTIPLE_OS)
```
ilia1243, line 49: New restriction. When migrating of OS, it is necessary to move all customizations to OS family specific sections.

shmo1218 (approved):
The doc file is reviewed.

Imadzuma (approved):

kubemarine/packages.py
koryaga: let's leave as is

koryaga (approved):

PR #287, opened 2022-11-21
Update check_iaas.py

nikhil1697:
### Description
Check_iaas should provide an warning if different OS versions are installed on nodes

### Solution
Adding a new check in the check_iaas procedure

### Test Cases

**TestCase 1**

Test Configuration:
Initial set-up:
node1: Ubuntu 22.04
node2: Ubuntu 20.04
node3: Ubuntu 20.04

Steps:

1. kubemarine check_iaas

Results:
 k8s-stack.sdntest.testcluster.com DEBUG Host 10.102.0.67 running "ubuntu 22.04"
 System     WARN    008  Distibutive ................................ Detected OS versions are not similar
          HINT:
               Install similar verions of OS on all nodes

image

### Checklist
- [ ] Integration CI passed

### DIFF
--- a/kubemarine/procedures/check_iaas.py
+++ b/kubemarine/procedures/check_iaas.py
@@ -258,6 +258,12 @@ def system_distributive(cluster):
             else:
                 detected_supported_os.append(detected_os)
                 cluster.log.debug('Host %s running \"%s\"' % (address, detected_os))
+                result = all(versions == detected_os[0] for versions in detected_os)
+                if (result):
+                    print ("Detected versions are similar")
+                else:
+                    raise TestWarn ("Detected OS versions are not similar",
+                    hint="Install similar verions of OS on all nodes")
 
         detected_supported_os = list(set(detected_supported_os))
         detected_unsupported_os = list(set(detected_unsupported_os))


kubemarine/procedures/check_iaas.py
```diff
@@ -258,6 +258,12 @@ def system_distributive(cluster):
             else:
                 detected_supported_os.append(detected_os)
                 cluster.log.debug('Host %s running \"%s\"' % (address, detected_os))
+                result = all(versions == detected_os[0] for versions in detected_os)
```
ilia1243, line 261: `result` is always false. Even for cluster consisted of the only node. Also, please firstly check that all OS are supported and  raise failure otherwise, and only if all OS are supported, check that they are equal.

ilia1243 (changes requested):

kubemarine/procedures/check_iaas.py
```diff
@@ -242,11 +242,12 @@ def system_distributive(cluster):
         detected_supported_os = []
         detected_unsupported_version = []
         supported_versions = []
+        detected_versions_compatability = []
```
ilia1243, line 245: It is simpler to use `set()`. In this case it is enough to add to the set and finally check that len() == 1.

kubemarine/procedures/check_iaas.py
```diff
@@ -257,7 +258,10 @@ def system_distributive(cluster):
                 cluster.log.error('Host %s running unknown OS family \"%s\"' % (address, detected_os))
             else:
                 detected_supported_os.append(detected_os)
-                cluster.log.debug('Host %s running \"%s\"' % (address, detected_os))
+                cluster.log.debug('Host %s running \"%s\"' % (address, detected_os)) 
+                detected_versions_compatability.append(detected_os)
```
ilia1243, line 262: `detected_os` is made from os.name and os.version. But in fact os.name can be different but os.family be the same. In Kubemarine, we consider nodes with the same os.family and os.version effectively the same. os.name has only informational sense. I suggest to postpose the ticket till #278 . cluster.get_os_identifiers() is introduced there and can be used.

kubemarine/procedures/check_iaas.py
```diff
@@ -272,6 +276,11 @@ def system_distributive(cluster):
             raise TestFailure("Unsupported version: %s" % ", ".join(detected_unsupported_version),
                               hint="Reinstall the OS on the host to one of the supported versions: %s" % \
                                       ", ".join(supported_versions))
+        if (result):
+            print ("Detected versions are similar")
```
ilia1243, line 280: use cluster.log

kubemarine/procedures/check_iaas.py
```diff
@@ -272,6 +276,11 @@ def system_distributive(cluster):
...
+        else:
+            raise TestWarn ("Detected OS versions are not similar",
+            hint="Install similar verions of OS on all nodes")
```
ilia1243, line 283: Probably add information to the warn message, what unique pairs (OS family, OS version) are installed.

kubemarine/procedures/check_iaas.py
nikhil1697: okay
Used same logic which is used for ticket #278 in check_paas to check the os versions now

kubemarine/procedures/check_iaas.py
```diff
@@ -272,7 +272,15 @@ def system_distributive(cluster):
...
                               hint="Reinstall the OS on the host to one of the supported versions: %s" % \
                                       ", ".join(supported_versions))
-
+        
+        os_ids = cluster.get_os_identifiers()
+        different_os = set(os_ids.values())
+        if len(different_os) > 1:
+            cluster.log.warning(
+                f"Nodes have different OS families or versions, packages versions cannot be checked. "
+                f"List of (OS family, version): {list(different_os)}")
```
ilia1243, line 281: This task `system.distributive` has nothing about packages, so the warning message is misleading.

kubemarine/procedures/check_iaas.py
```diff
@@ -272,7 +272,15 @@ def system_distributive(cluster):
...
+            raise TestFailure(f"Nodes have different OS families or versions")
```
ilia1243, line 282: The description says: `Check_iaas should provide an warning if different OS versions are installed on nodes`, but the test is failed. So, warning or failure?

nikhil1697: Changed it to TestWarn

kubemarine/procedures/check_iaas.py
nikhil1697: Removed packages versions

ilia1243 (approved):

koryaga (approved):

PR #297, opened 2022-11-29
[MANOPD-81952] - use internal address if address is not defined

Imadzuma:
### Description
We have optional field address for, but if it's actually undefined, the exception occurs.

Fixes # (issue)

### Solution
If address is not defined, internal address is used instead of it.
* 

### How to apply
* Any procedure without address field on some nodes in cluster.yaml

### Test Cases

**TestCase 1**

Test Configuration:

- Hardware: 
- OS: 
- Inventory: 

Steps:

1. 

Results:

| Before | After |
| ------ | ------ |
|  |  |

### Checklist
- [ ] Integration CI passed

#### Unit tests


kubemarine/core/defaults.py
```diff
@@ -93,6 +93,10 @@ def apply_defaults(inventory, cluster):
             raise Exception('Node name \"%s\" contains invalid characters. A DNS-1123 subdomain must consist of lower '
                             'case alphanumeric characters, \'-\' or \'.\'' % node_name)
 
+        # Use internal address if address is not defined
+        if node.get('address') is None and node.get("internal_address") is not None:
+            cluster.inventory["nodes"][i]["address"] = node.get("internal_address")
```
ilia1243, line 98: Presence of address property influences the behaviour of Kubemarine. For example, see defaults.append_controlplain and system.generate_etc_hosts_config. Won't it be safe if we just check presence of the property where necessary?
Also I suggest to develop unit test which ensure that enrichment is successful without address property.
If you still want to assign address = internal_address and have explanation why it is safe, this makes many places in code doing unnecessary checks (for example, the same places in defaults and system mentioned).

Imadzuma: Yes, I see. I decided not to fill this value from internal address during enrichment. 
I've checked all places, there 'address' appears and found two places, where we don't have such check. Those places were fixed.

ilia1243: Thank you, and what about this
> Also I suggest to develop unit test which ensure that enrichment is successful without address property.
Based on the implementation, I conclude that tests of generation of final and finalized inventory are also desirable.

Imadzuma: Test was added

test/unit/test_inventory.py
```diff
@@ -96,5 +96,53 @@ def test_roles_in_inventory(self):
...
+            'nodes': [copy.deepcopy(inventory['nodes'][0])]
+        }
+
+        # Remove node inventory
+        context = demo.create_silent_context(procedure='remove_node')
+        cluster = demo.new_cluster(inventory, procedure_inventory=procedure_inventory, context=context)
+        for node in cluster.inventory['nodes']:
+            self.assertNotIn('address', node)
+
+        final_inventory = utils.get_final_inventory(cluster)
```
ilia1243, line 126: It is preferable to pass original inventory to the utils.get_final_inventory(cluster, inventory), because it is always passed that one during real Kubemarine work.

ilia1243: Hmm, actually there are two cases: 1) generation of ansible inventory, passed without second argument 2) recreation of the inventory, passed with the second argument. Ok then, let's leave it this way.

ilia1243 (approved):

koryaga:
@Imadzuma  pls also update our README example and remove extra parameters

PR #300, opened 2022-12-02
[MANOPD-82181] - support string plugin.install values

Imadzuma:
### Description
Because of using jinja we have to support string true/false values. But for `plugins.<name>.install` we support only bool types.
As result  `plugins.<name>.install='false'` don't skip this plugin installation or `plugins.nginx-ingress-controller.install='true'` throws exception during cert renew procedure for ingress-nginx.

### Solution
`true_or_false` function, what is used in calico plugin, was moved to utils. This function was used in conditional statements with `plugins.<name>.install`

### Test Cases

**TestCase 1**

Steps:

1.  Add `plugins.kubernetes-dashboard.install='false'` to cluster.yaml;
2. Run installation procedure;

Results:

| Before | After |
| ------ | ------ |
| kubernetes-dashboard is installed to cluster | kubernetes-dashboard isn't installed |

**TestCase 1**

Steps:

1.  Add `plugins.nginx-ingress-controller.install='true'` to cluster.yaml;
2. Run cert renew procedure with new cert for ingress-nginx;

Results:

| Before | After |
| ------ | ------ |
| Procedure fails with exception | cert is updated |

### Checklist
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/kubemarine/core/utils.py
+++ b/kubemarine/core/utils.py
@@ -116,7 +116,7 @@ def make_ansible_inventory(location, cluster):
         if inventory.get(group) is not None:
             for service_name, service_configs in inventory[group].items():
                 # write to inventory only plugins, which will be installed
-                if group != 'plugins' or service_configs.get('install', False) is True:
+                if group != 'plugins' or true_or_false(service_configs.get('install', False)) == 'true':
 
                     config['cluster:vars'].append('\n# %s.%s' % (group, service_name))
 
--- a/kubemarine/core/utils.py
+++ b/kubemarine/core/utils.py
@@ -295,6 +295,21 @@ def load_yaml(filepath) -> dict:
         do_fail(f"Failed to load {filepath}", exc)
 
 
+def true_or_false(value):
+    """
+    The method check string and boolean value
+    :param value: Value that should be checked
+    """
+    input_string = str(value)
+    if input_string in ['true', 'True', 'TRUE']:
+        result = "true"
+    elif input_string in ['false', 'False', 'FALSE']:
+        result = "false"
+    else:
+        result = "undefined"
+    return result
+
+
 class ClusterStorage:
     """
     File preservation:
--- a/kubemarine/plugins/__init__.py
+++ b/kubemarine/plugins/__init__.py
@@ -135,7 +135,8 @@ def install(cluster, plugins=None):
     plugins_queue = []
     max_priority = 0
     for plugin_name, plugin_item in plugins.items():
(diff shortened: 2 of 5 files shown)


kubemarine/plugins/__init__.py
```diff
@@ -135,7 +135,8 @@ def install(cluster, plugins=None):
     plugins_queue = []
     max_priority = 0
     for plugin_name, plugin_item in plugins.items():
-        if plugin_item.get("install", False) and plugin_item.get("installation", {}).get('procedures') is not None:
+        if utils.true_or_false(plugin_item.get("install", False)) == 'true' and \
```
ilia1243, line 138: Yet another places in
* utils.make_ansible_inventory
* calico.py enrich_inventory
* few places in check_paas

Probably also add method like plugins.is_to_be_installed(inventory, plugin_name)

kubemarine/core/utils.py
```diff
@@ -295,6 +295,21 @@ def load_yaml(filepath) -> dict:
         do_fail(f"Failed to load {filepath}", exc)
 
 
+def true_or_false(value):
```
ilia1243, line 298: I would also suggest to consider:
* SystemdCgroup in containerd.configure
* enabled in coredns.generate_nested_sections

kubemarine/plugins/__init__.py
Imadzuma: `plugins.is_to_be_installed(inventory, plugin_name)` is not so good, because during concertation we can give 3 variants: true, false and unsupported (value that can't be converted).  
Missed places were fixed

kubemarine/core/utils.py
Imadzuma: In this PR I fix only plugins.install. 
Of course this problem appears for other values, but we should design a generic approach to fix all places.
Ticket number for it: MANOPD-82348

alexarefev (approved):

n549 (approved):

kubemarine/plugins/__init__.py
ilia1243: > we can give 3 variants: true, false and unsupported (value that can't be converted).

and actually this is a good point. In different places of code we consider "undefined" as False (for example, in plugins.install) and as True (for example in nginx_ingress.py) even for the same plugin. I suggest to throw exception for "undefined" during enrichment. (plugins.verify_inventory or better in some new procedure which will be run before any plugin enrichment). After this, it would be still possible to introduce some new method like is_to_be_installed

PR #309, opened 2022-12-09
[MANOPD-82746] containerd.io versions fix

alexarefev:
### Description
* The official repositories for Ubuntu 20.04 and Ubuntu 22.04 have different set of `containerd.io` versions. The Kubernetes installation on Ubuntu 22.04 fails if `containerd.io` version set to 1.4.* whereas the installation on Ubuntu 20.04 goes smoothly

### Solution
*  Change the `compatibility_map` in `globals.yaml` and set it to the version that presents in both repositories, since we don't have separated versions for Ubuntu 22.04 and Ubuntu 20.04

### How to apply
* Those changes are not compatible with artifacts that were created by previous `KubeMarine` versions, in case of using the Ubuntu 22.04. For instance, `cluster_finalized.yzml` can't be used for `add_node` operation.
* The following procedure doesn't work on clusters that have several `containerd.io` version:
[Migrate kubemarine](documentation/Maintenance.md#kubemarine-migration-procedure)

### Test Cases

**TestCase 1**

Test Configuration:

- Hardware: 4CPU/4GB
- OS: Ubuntu 22.04
- Inventory: AllInOne

Steps:

1. Run installation procedure. The `cluster.yaml` has  `services.cri.containerRuntime: docker` option

Results:

| Before | After |
| ------ | ------ |
| Fails on task `prepare.cri` | Success |

### Checklist
- [ ] Integration CI passed

### DIFF
--- a/documentation/Installation.md
+++ b/documentation/Installation.md
@@ -2135,9 +2135,11 @@ services:
 For detailed description of the parameters, see [Installation without Internet Resources](#installation-without-internet-resources).
 For more information about Docker daemon parameters, refer to the official docker configuration file documentation at [https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file).
 
-**Note**: After applying the parameters, the docker is restarted on all nodes in the cluster.
+**Notes**: 
+* After applying the parameters, the docker is restarted on all nodes in the cluster.
+* Do not omit the `containerRuntime` parameter in cluster.yaml if you include `dockerConfig` or `containerdConfig` in `cri` section
 
-**Note**: Do not omit the `containerRuntime` parameter in cluster.yaml if you include `dockerConfig` or `containerdConfig` in `cri` section
+**Warning:** Kubernetes v1.24 and higher do not work with `docker` as a `cri`
 
 #### modprobe
 
--- a/kubemarine/patches/p1_upgrade_containerdio_version.py
+++ b/kubemarine/patches/p1_upgrade_containerdio_version.py
@@ -1,0 +1,60 @@
+# Copyright 2021-2022 NetCracker Technology Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from kubemarine.core import flow
+from kubemarine.core.action import Action
+from kubemarine.core.patch import Patch
+from kubemarine.core.resources import DynamicResources
+from kubemarine import cri
+from kubemarine import packages
+
+
+class TheAction(Action):
+    def __init__(self):
+        super().__init__("Upgrade containerd.io version")
+
+    def run(self, res: DynamicResources):
+        cluster = res.cluster()
+
+        # check the OS family
+        if cluster.get_os_family() == 'debian':
+            # check CRI
+            if cluster.inventory['services']['cri']['containerRuntime']:
+                # get installed version
+                group = cluster.nodes['all'].get_accessible_nodes()
+                result = packages.detect_installed_packages_version_groups(group, 'containerd.io')
+                version = list(result['containerd.io'])[0].split('=')[1]
+                if '.'.join(version.split('.')[:2]) == '1.4':
+                    group.call(cri.install)
+                else:
+                    cluster.log.debug("Cluster has applicable \"containerd.io\" version")
+            else:
+                cluster.log.debug("Cluster has \"Containerd\" as a CRI, so the \"containerd.io\" upgrade is not needed")
+        else:
+            cluster.log.debug("Cluster has different OS family, so the \"containerd.io\" upgrade is not needed")
+
+
+class UpgrateContainerdioVersion(Patch):
+    def __init__(self):
+        super().__init__("upgrade_containerdio_version")
+
+    @property
+    def action(self) -> Action:
+        return TheAction()
+
+    @property
+    def description(self) -> str:
+        return """\
+Upgrade containerd.io to v1.5.* for kubernetes clusters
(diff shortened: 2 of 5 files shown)


documentation/Installation.md
```diff
@@ -2135,9 +2135,11 @@ services:
 For detailed description of the parameters, see [Installation without Internet Resources](#installation-without-internet-resources).
 For more information about Docker daemon parameters, refer to the official docker configuration file documentation at [https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file).
 
-**Note**: After applying the parameters, the docker is restarted on all nodes in the cluster.
+**Notes**: 
+* After applying the parameters, the docker is restarted on all nodes in the cluster.
+* Do not omit the `containerRuntime` parameter in cluster.yaml if you include `dockerConfig` or `containerdConfig` in `cri` section
 
-**Note**: Do not omit the `containerRuntime` parameter in cluster.yaml if you include `dockerConfig` or `containerdConfig` in `cri` section
+**Warning:** Kubernetes v1.24 and higher do not work with `docker` as a `cri`
```
ilia1243, line 2142: Probably we can ensure this in Kubemarine code? We can explicitly fail during attempt to install / upgrade cluster with docker and Kubernetes >= v1.24

kubemarine/patches/p1_upgrade_containerdio_version.py
```diff
@@ -0,0 +1,60 @@
...
+    def __init__(self):
+        super().__init__("Upgrade containerd.io version")
+
+    def run(self, res: DynamicResources):
+        cluster = res.cluster()
+
+        # check the OS family
+        if cluster.get_os_family() == 'debian':
+            # check CRI
+            if cluster.inventory['services']['cri']['containerRuntime']:
```
ilia1243, line 33: Forget to compare with == 'docker'

kubemarine/patches/p1_upgrade_containerdio_version.py
```diff
@@ -0,0 +1,60 @@
...
+                # get installed version
+                group = cluster.nodes['all'].get_accessible_nodes()
+                result = packages.detect_installed_packages_version_groups(group, 'containerd.io')
+                version = list(result['containerd.io'])[0].split('=')[1]
+                if '.'.join(version.split('.')[:2]) == '1.4':
```
ilia1243, line 38: This is a random behavior in case different nodes have different versions. Probably make it more robust and raise error / log warning / install anyway in such case.

ilia1243 (changes requested):

kubemarine/patches/p1_upgrade_containerdio_version.py
alexarefev: This random behavior inherits from the package version detection method which was not perfectly designed.

ilia1243: Could you specify problems in the package version detection method? For me, it does its task. It returns all versions from all nodes, including the case when the versions are different.

ilia1243: Actually current implementation has one more problem. docker might be not installed on balancers, but you query balancers. Since you take the random result, it might be resolved to "not installed"

documentation/Installation.md
alexarefev: Yes, I agree. It must be implemented

kubemarine/patches/p1_upgrade_containerdio_version.py
alexarefev: The `cluster.yaml` structure implies only one version of any package for all nodes in the cluster. Therefore several version for the same package must raise an exception.

ilia1243: > Therefore several versions for the same package must raise an exception.

Could you clarify, is this related to the design of detect_installed_packages_version_groups? The method is used in particular by check_paas. So processing of the result is delegated to the caller by design.
If you are about my initial proposal in this thread, then agree, raising of error might be one of the solutions. Though, you might additionally check if cache_package=False is configured, because it allows different package versions in the cluster.

alexarefev: The `cache_package=False` option is just a bypass. I'm talking about a comprehensive solution which doesn't exist in a core, but you expect it in a patch.

ilia1243: Comprehensive solution exists in core. All users of the method detect_installed_packages_version_groups take into account probably different package versions. If you need special behavior only for single patch and do not want to implement it directly in the patch, you can implement it in some other place.

ilia1243:
> Those changes are not compatible with artifacts that were created by previous KubeMarine versions, in case of using the Ubuntu 22.04. For instance, cluster_finalized.yzml can't be used for add_node operation.

cluster_finalized.yaml cached `1.4.<exact>` versions of containerd.io and cluster worked well on those versions. During add_node, exactly the same `1.4.<exact>` version will be installed, and the cluster will still work well. I do not see compatibility problems here.

alexarefev:
> > Those changes are not compatible with artifacts that were created by previous KubeMarine versions, in case of using the Ubuntu 22.04. For instance, cluster_finalized.yzml can't be used for add_node operation.
> 
> cluster_finalized.yaml cached `1.4.<exact>` version of containerd.io and cluster worked well on that version. During add_node, exactly the same `1.4.<exact>` version will be installed on new nodes, and the cluster will still work well. I do not see compatibility problems here.

In this case the whole PR does not make any sense. But for Ubuntu 22.04 the list of available packages is the following:
```
$ apt list -a containerd.io
Listing... Done
containerd.io/jammy 1.6.14-1 amd64
containerd.io/jammy 1.6.13-1 amd64
containerd.io/jammy 1.6.12-1 amd64
containerd.io/jammy 1.6.11-1 amd64
containerd.io/jammy 1.6.10-1 amd64
containerd.io/jammy 1.6.9-1 amd64
containerd.io/jammy 1.6.8-1 amd64
containerd.io/jammy 1.6.7-1 amd64
containerd.io/jammy 1.6.6-1 amd64
containerd.io/jammy 1.6.4-1 amd64
containerd.io/jammy 1.5.11-1 amd64
containerd.io/jammy 1.5.10-1 amd64
```

ilia1243:
> For Ubuntu 22.04 the list of available packages is the following:

So, previous KubeMarine version is not able to install the cluster at all, so the given sentence in How to Apply is again not relevant.

> In this case the whole PR does not make any sense

I did not indent to doubt in the goal of the whole PR, so I am not sure how you have made such conclusion.

PR #332, opened 2023-01-12
Cumulative points fixes

ilia1243:
### Description
* KubeMarine performs unnecessary reboot during `install` procedure in some cases especially is cluster is reinstalled.
* Cumulative points are not executed if the related task is skipped, and the system is not verified in such cases.

### Solution
* Remove unnecessary reboot schedule from disable swap.
* Fix packages changes detection for Ubuntu.
* Flow now always iterates over all tasks to proceed cumulative points, but only necessary tasks are finally run.
* Other refactoring, removed dynamic imports.

### Test Cases

**TestCase 1**

Test Configuration:

- Hardware: Any
- OS: Ubuntu
- Inventory: should contain non-empty `services.packages.install` section.

Steps:

1. Run `install` procedure twice.

Results:

| Before | After |
| ------ | ------ |
| Both runs reboot the nodes | The second run should not reboot the nodes |

**TestCase 2**

Test Configuration:

- Hardware: Any
- OS: Any
- Inventory: Any

Steps:

1. Install and/or enable `firewalld`.
1. Run `install --tasks prepare.system.disable_firewalld`.

Results:

| Before | After |
| ------ | ------ |
| Nodes are not rebooted and the system is not verified after reboot | The nodes are rebooted and the system is verified |

### Checklist
- [x] Integration CI passed

#### Unit tests
test_flow.py - added new tests and changed the existing to adopt to the changed methods signatures.

### DIFF
--- a/kubemarine/core/flow.py
+++ b/kubemarine/core/flow.py
@@ -144,7 +144,7 @@ def run_tasks(resources: res.DynamicResources, tasks, cumulative_points=None, ta
         return
 
     init_tasks_flow(cluster)
-    run_flow(filtered_tasks, cluster, cumulative_points)
+    run_flow(tasks, final_list, cluster, cumulative_points, [])
 
 
 def create_empty_context(args: dict = None, procedure: str = None):
@@ -198,6 +198,7 @@ def _filter_flow_internal(tasks, tasks_filter: List[List[str]], excluded_tasks: 
 
     for task_name, task in tasks.items():
         __task_path = _task_path + [task_name]
+        __task_name = ".".join(__task_path)
 
         allowed = True
         # if task_filter is not empty - smb specified filter argument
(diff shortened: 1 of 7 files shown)


kubemarine/core/flow.py
```diff
@@ -146,7 +144,7 @@ def run_tasks(resources: res.DynamicResources, tasks, cumulative_points=None, ta
         return
 
     init_tasks_flow(cluster)
-    run_flow(filtered_tasks, cluster, cumulative_points)
+    run_flow(tasks, final_list, cluster, cumulative_points, [])
```
ilia1243, line 147: I suggest to always run scheduled cumulative points at the right moment. To see the advantages, refer to `Test Case 2` from the description.

It imply that KubeMarine runs will be slower in some cases, especially if the task is run over the existing cluster and graceful reboot is required.

On the other hand, only scheduled reboot is affected, and it should be scheduled only when necessary and only when the changes in the system are detected. See another part of the PR.

dmyar21 (approved):

theboringstuff (approved):

koryaga (approved):

kubemarine/core/flow.py
koryaga: ok. @OleksandrIaresko  pls check the performance impact

PR #346, opened 2023-01-27
MANOPD-84031 Input files encoding rework and non-ASCII upload

### LINKED ISSUE #342: Cannot upload io.StringIO with non-ASCII characters
Steps to reproduce:
1. Create template files/test_cm.yaml.j2
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-cm
data:
  cluster_name: "{{ cluster_name }}"
  unicode_symbol: "😀"
```
2. Create custom plugin in inventory
```yaml
plugins:
  ...
  my_plugin:
    install: true
    installation:
      procedures:
        - template:
            source: files/test_cm.yaml.j2
```
3. Run `install --tasks deploy.plugins`

AR: fails with `size mismatch in put!  121 != 118`, see https://github.com/paramiko/paramiko/issues/1133

The problem is likely actual for any upload of io.StringIO instance.

Possible solution:

When uploading of any text, manually encode it with 'utf-8' and upload as io.BytesIO

If the text is obtained from file on Windows deployer, it will need to decode it first with the same 'utf-8' to not lose the utf-8 symbols.

In other words, it is necessary to introduce new requirement for encoding of input text files and **always open for reading and writing with 'utf-8'** and always upload as io.BytesIO.

Prototype: https://github.com/Netcracker/KubeMarine/commit/eedfbf916b999b7eeb343d58a34d0c725e8bcc14

ilia1243:
### Description
There are many places in KubeMarine where there can be an attempt to upload non-ASCII characters. This does not work because of issue https://github.com/paramiko/paramiko/issues/1133
The problem is especially relevant for custom template plugins.

Fixes #342

### Solution
* `NodeGroup.put` now uploads data only as stream of bytes, not characters.
* New requirement for all input text files to be strictly in utf-8 encoding.
    * This is reflected in new utils.py methods which now always open in utf-8.
    * Any internal or external text file should be opened using new methods.
* New approach automatically leads to CRLF -> LF transformation of line endings on Windows deployer for any input text file (internal or external) except external `config` files for plugins with `do_render=False`.
* Removed `binary` parameter for thirdparties as uploading from filesystem is not supported for external thirdparties.
* Minor fix in `services.packages.package_manager.repositories`: now it allows input as path to file for both apt and yum package managers.

### Breaking changes
If `services.packages.package_manager.repositories` for Debian-based cluster was passed as string, it will stop working as the string is now considered as path to file. No longer working configuration:
```yaml
services:
  packages:
    package_manager:
      repositories: |
        deb http://archive.ubuntu.com/ubuntu focal main restricted
        deb http://archive.ubuntu.com/ubuntu focal-updates main restricted
```
New working configuration:
```yaml
services:
  packages:
    package_manager:
      repositories: files/predefined.list
```
where files/predefined.list contains the repositories.

### Test Cases

See https://github.com/Netcracker/KubeMarine/issues/342

### Checklist
- [x] Integration CI passed

#### Unit tests
test_demo.py and demo.py to adopt to new upload behaviour.

### DIFF
--- a/kubemarine/apt.py
+++ b/kubemarine/apt.py
@@ -16,6 +16,7 @@ import io
 
 import fabric
 
+from kubemarine.core import utils
 from kubemarine.core.group import NodeGroupResult, NodeGroup
 
 DEBIAN_HEADERS = 'DEBIAN_FRONTEND=noninteractive '
@@ -49,7 +50,7 @@ def create_repo_file(group, repo_data, repo_file):
     if isinstance(repo_data, list):
         repo_data_str = "\n".join(repo_data) + "\n"
     else:
-        repo_data_str = str(repo_data)
+        repo_data_str = utils.read_external(repo_data)
     group.put(io.StringIO(repo_data_str), repo_file, sudo=True)
 
 
--- a/documentation/Installation.md
+++ b/documentation/Installation.md
@@ -159,6 +159,7 @@ System clock should be synchronized the same way as for Cluster nodes system clo
 
 There are the following restrictions when deploying from Windows:
 * [ansible](#ansible) plugin procedures are not supported.
+* All KubeMarine input text files must be in utf-8 encoding.
 
 ## Prerequisites for Cluster Nodes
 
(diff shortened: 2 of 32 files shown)


kubemarine/apt.py
```diff
@@ -47,7 +48,7 @@ def create_repo_file(group, repo_data, repo_file):
     if isinstance(repo_data, list):
         repo_data_str = "\n".join(repo_data) + "\n"
     else:
-        repo_data_str = str(repo_data)
+        repo_data_str = utils.read_external(repo_data)
```
ilia1243, line 51: This is an incompatibly change. Previously yum considered string as path to file with repositories content (see yum.py), and apt considered as the repositories content itself. Now both consider the string as path to file.

Imadzuma (approved):

shmo1218 (approved):

shmo1218:
The doc review is done.

kubemarine/apt.py
koryaga: Pls add **Breaking change:** section in PR

ilia1243: Done

koryaga:
@alku89  pls note about Breaking changes

PR #347, opened 2023-01-30
[MANOPD-84029] Status of resources

alexarefev:
### Description
Installation/upgrade procedure doesn't fail on `deploy.plugins` task if the retries are exceeded for the following resources:
* `DaemonSet`
* `Deployment`
* `StatefulSet`
* `ReplicaSet`

### Solution
* Add exception in `expect_` methods in `plugins` module

### How to apply
Not applicable

### Test Cases

**TestCase 1**
Check if the fix works

Test Configuration:

- Hardware: 4CPU/4GB
- OS: Ubuntu 20.04
- Inventory: MiniHA

Steps:

1. Run installation procedure. At least one predefined plugin must be enabled and it should not have enough time to run its main resource (one of the following: `DaemonSet`, `Deployment`, `StatefulSet`, `ReplicaSet`)

Results:

| Before | After |
| ------ | ------ |
| Success | Fail with exception |

### Checklist
- [x] Integration CI passed

#### Unit tests

### DIFF
--- a/kubemarine/resources/configurations/globals.yaml
+++ b/kubemarine/resources/configurations/globals.yaml
@@ -52,7 +52,7 @@ pods:
       retries: 30
     plugins:
       timeout: 5
-      retries: 30
+      retries: 45
   critical_states:
     - Error
     - ErrImagePull
--- a/kubemarine/resources/configurations/globals.yaml
+++ b/kubemarine/resources/configurations/globals.yaml
@@ -52,7 +52,7 @@ pods:
       retries: 30
     plugins:
       timeout: 5
-      retries: 360
+      retries: 30
   critical_states:
     - Error
     - ErrImagePull
--- a/kubemarine/resources/configurations/globals.yaml
+++ b/kubemarine/resources/configurations/globals.yaml
@@ -52,7 +52,7 @@ pods:
       retries: 30
     plugins:
       timeout: 5
-      retries: 45
+      retries: 360
   critical_states:
     - Error
     - ErrImagePull
--- a/kubemarine/plugins/__init__.py
+++ b/kubemarine/plugins/__init__.py
@@ -253,6 +253,7 @@ def expect_replicaset(cluster: KubernetesCluster,
 
     raise Exception('In the expected time, the ReplicaSets did not become ready')
 
+
 def expect_statefulset(cluster: KubernetesCluster,
                        statefulsets_names: List[str] or List[Dict[str, str]],
                        timeout: int = None,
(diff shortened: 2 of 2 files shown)


kubemarine/resources/configurations/globals.yaml
```diff
@@ -52,7 +52,7 @@ pods:
       retries: 30
     plugins:
       timeout: 5
-      retries: 30
+      retries: 45
```
n549, line 55: Probably it's better to increase number of retries more significantly, for example, to 360 (so that the waiting time was up to  a half an hour). This should cover most cases with long starting of pods (calico-node, for example).

alexarefev: Done

n549 (approved):

andrewluckyguy (approved):
Verified by QA for 'expect_daemonset' method

kubemarine/resources/configurations/globals.yaml
```diff
@@ -52,7 +52,7 @@ pods:
       retries: 30
     plugins:
       timeout: 5
-      retries: 30
+      retries: 360
```
ilia1243, line 55: Just a point to think of. Do we expect different time to become ready for different OOB plugins? If yes, we could specify timeout or retries in the defaults.yaml for the `expect` installation procedure directly, for example:
```yaml
        - expect:
            pods:
              list:
                - coredns
                - calico-kube-controllers
                - calico-node
              retries: 360
```

ilia1243: Also I'd like to warn that unsuccessful installation will become not friendly, as it will hang the Kubemarine run for half an hour.

kubemarine/resources/configurations/globals.yaml
ilia1243: If we expect different time to become ready for different OOB plugins, then we could specify timeout or retries in the defaults.yaml for the `expect` installation procedure directly, for example:
```yaml
        - expect:
            pods:
              list:
                - coredns
                - calico-kube-controllers
                - calico-node
              retries: 360
```

ilia1243: Though, even in this case, I'd like to warn that unsuccessful installation will become not friendly, as it will hang the Kubemarine run for half an hour, even if it is technically clear that the pods will never be ready after very smaller time period.

koryaga: Let's have 45 for now. 360 looks too ambisious.

koryaga (approved):

ilia1243 (approved):

PR #367, opened 2023-02-22
[PSUPCLPL-11680] adjust plugins and nodes timeouts

n549:
### Description
Need to do the following improvements:
1. Add a separate timeout for expectation of daemonsets/deployments/replicasets/statefulsets readyness (now we have only one timeout pods.expect.plugins which is used for all of them and for pods).
2. Set higher default timeout for expectation of calico-node pods.
3. Add more sense in exception messages when daemonsets/deployments/replicasets/statefulsets/pods have not got ready within given timeouts (how to override default values of timeouts/retries).
4. Make some global timeouts configurable (nodes.ready, nodes.boot.defaults.timeout).

Fixes # (issue)
PSUPCLPL-11680
PSUPCLPL-11836
PSUPCLPL-11876

### Solution
1. New global parameter is introduced:
```
deployments:
  expect:
    timeout: 5
    retries: 30
```
which defines default values for `timeout` and `retries` in expectation of deployments/daemonsets/replicasets/statefulsets.

2. Increased number of retries in `expect.pods` for calico-node in defaults.yaml.

3. Exception messages in expect_* are updated.

4. New section `globals: {}` in cluster.yaml is introduced. It has the similar structure to `globals.yaml`, but user can override only these parameters. Now these are allowed:
```
globals:
  nodes:
    ready:
      timeout: 10
      retries: 50
    boot:
      timeout: 1000
```
Other parameters (for example, deployment or pods timeouts) can also be allowed if necessary.

Log messages in corresponding procedures are also updated with information about timeouts.

### How to apply
-

### Test Cases
1. Deploy k8s cluster with new parameters in cluster.yaml, for example:
```
node_defaults:
  username: "ubuntu"
  keyfile: "/home/nat/.ssh/shift-test.key"
  ready_timeout: 3
  ready_retries: 57
  boot_timeout: 876
```
ER: cluster is deployed successfully.

2. Try to add a node to the cluster from step 1.
ER: the node is added successfully, in debug.log there is a message like:
```
DEBUG [__init__.wait_for_nodes] Waiting for new kubernetes nodes to become ready, 57 retries every 3 seconds
```

3. Try to reboot nodes in the cluster.
ER: the nodes rebooted successfully, in debug log there is a message like:
```
VERBOSE [group._await_rebooted_nodes] Trying to connect to nodes, timeout is 876 seconds...
```

### Checklist
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/kubemarine/core/group.py
+++ b/kubemarine/core/group.py
@@ -630,8 +630,7 @@ class NodeGroup:
     def _await_rebooted_nodes(self, timeout=None, initial_boot_history: NodeGroupResult = None) -> _HostToResult:
 
         if timeout is None:
-            # boot_timeout is set for all nodes, so we can take any node's boot_timeout
-            timeout = self.cluster.inventory['nodes'][0].get('boot_timeout', self.cluster.globals['nodes']['boot']['defaults']['timeout'])
+            timeout = int(self.cluster.inventory['globals']['nodes']['boot']['timeout'])
 
         delay_period = self.cluster.globals['nodes']['boot']['defaults']['delay_period']
 
@@ -680,7 +679,7 @@ class NodeGroup:
                 time.sleep(delay_period - attempt_time)
 
         if left_nodes:
-            self.cluster.log.verbose("Failed to wait for boot of nodes %s. Try to increase boot_timeout parameter in node_defaults: https://github.com/Netcracker/KubeMarine/blob/main/documentation/Installation.md#node_defaults" % list(left_nodes.keys()))
+            self.cluster.log.verbose("Failed to wait for boot of nodes %s" % list(left_nodes.keys()))
         else:
             self.cluster.log.verbose("All nodes are online now")
 
--- a/kubemarine/core/group.py
+++ b/kubemarine/core/group.py
@@ -630,7 +630,8 @@ class NodeGroup:
     def _await_rebooted_nodes(self, timeout=None, initial_boot_history: NodeGroupResult = None) -> _HostToResult:
 
         if timeout is None:
-            timeout = self.cluster.globals['nodes']['boot']['defaults']['timeout']
+            # boot_timeout is set for all nodes, so we can take any node's boot_timeout
+            timeout = self.cluster.inventory['nodes'][0].get('boot_timeout', self.cluster.globals['nodes']['boot']['defaults']['timeout'])
 
         delay_period = self.cluster.globals['nodes']['boot']['defaults']['delay_period']
 
@@ -642,6 +643,8 @@ class NodeGroup:
         left_nodes = self.nodes
         results: _HostToResult = {}
         time_start = datetime.now()
+
+        self.cluster.log.verbose("Trying to connect to nodes, timeout is %s seconds..." % timeout)
 
         # each connection has timeout, so the only we need is to repeat connecting attempts
         # during specified number of seconds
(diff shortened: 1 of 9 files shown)


kubemarine/core/group.py
```diff
@@ -677,7 +680,7 @@ def _await_rebooted_nodes(self, timeout=None, initial_boot_history: NodeGroupRes
                 time.sleep(delay_period - attempt_time)
 
         if left_nodes:
-            self.cluster.log.verbose("Failed to wait for boot of nodes %s." % list(left_nodes.keys()))
+            self.cluster.log.verbose("Failed to wait for boot of nodes %s. Try to increase boot_timeout parameter in node_defaults: https://github.com/Netcracker/KubeMarine/blob/main/documentation/Installation.md#node_defaults" % list(left_nodes.keys()))
```
ilia1243, line 683: This is misleading when removing of offline nodes. Such check have different timeout of 30 seconds and is currently not configurable.

kubemarine/plugins/__init__.py
```diff
@@ -439,7 +439,7 @@ def expect_pods(cluster, pods, namespace=None, timeout=None, retries=None,
             cluster.log.debug(running_pods_stdout)
             time.sleep(timeout)
 
-    raise Exception('In the expected time, the pods did not become ready')
+    raise Exception('In the expected time, the pods did not become ready. Try to increase number of retries in expect.pods: https://github.com/Netcracker/KubeMarine/blob/main/documentation/Installation.md#expect-pods')
```
ilia1243, line 442: This will be misleading if the method is invoked from `kubernetes.wait_for_any_pods`. In that case the timeout is not configurable.

kubemarine/core/group.py
n549: Yes, this message is removed to avoid confusion.

kubemarine/core/group.py
```diff
@@ -630,7 +630,8 @@ def wait_and_get_active_nodes(self, timeout=None) -> 'NodeGroup':
     def _await_rebooted_nodes(self, timeout=None, initial_boot_history: NodeGroupResult = None) -> _HostToResult:
 
         if timeout is None:
-            timeout = self.cluster.globals['nodes']['boot']['defaults']['timeout']
+            # boot_timeout is set for all nodes, so we can take any node's boot_timeout
+            timeout = self.cluster.inventory['nodes'][0].get('boot_timeout', self.cluster.globals['nodes']['boot']['defaults']['timeout'])
```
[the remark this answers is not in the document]
n549: New section `globals: {}` is added to cluster.yaml with the same structure as in globals.yaml for node readiness and reboot timeouts. It overrides values defined in globals.yaml for allowed parameters.

Parameter `timeout_download` can also be moved in `globals:`, but in a separate PR because it is out of scope of this PR.

kubemarine/plugins/__init__.py
n549: Removed for now. Maybe it's worth to consider making both timeouts `pods.expect.plugins` and `pods.expect.kubernetes` configurable via `globals:{}`.

kubemarine/core/group.py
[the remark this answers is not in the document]
n549: Issue on `timeout_download`: https://github.com/Netcracker/KubeMarine/issues/374

ilia1243 (approved):

koryaga (approved):

PR #371, opened 2023-03-01
[PSUPCLPL-11899] kubemarine/kubernetes/__init__.py

### LINKED ISSUE #363: KubeMarine should check if the 'services.kubeadm.networking.podSubnet' option is a valid IPv4 or IPv6 network.
I think, it should work.
There is a small addition. It's out of current scope, but connected to it.
KubeMarine should check if the 'services.kubeadm.networking.podSubnet' option is a valid IPv4 or IPv6 network.

_Originally posted by @alexarefev in https://github.com/Netcracker/KubeMarine/pull/360#pullrequestreview-1305098232_

nikhil1697:
### Solution
The code change first retrieves the podSubnet and serviceSubnet IP addresses from the inventory dictionary using dictionary get methods. It then attempts to create an IP network object using the ipaddress.ip_network method and the retrieved IP address.

If the IP address is not valid, an exception is raised and a ValueError is raised with a message indicating the invalid IP address. Otherwise, if the IP address is valid, the version of the IP network is checked to ensure it is either version 4 or 6.
Fixes #363 

### Test Cases

**TestCase 1**

Below ipv4 and ipv6 were used to check the improvement,

Valid IPv4 Subnets:                                                                          
192.168.0.0/16
10.0.0.0/8
172.16.0.0/12

Invalid IPv4 Subnets:
192.168.256.0/24 (invalid octet value)
10.0.0.0/33 (invalid prefix length)
172.31.0.0 (missing prefix length)

Valid IPv6 Subnets:

2001:db8: 1234::/48 (network address contains only 48 bits with room for 16 more bits to address hosts)
2001:db8:85a3::/64 (standard IPv6 subnet size)
2001:db8:85a3:0:0:8a2e:370:7334/128 (specifies a single IPv6 address with no room for hosts)

Invalid IPv6 Subnets:

(fe80:/64 (missing address value)
2001:db8:85a3::8a2e:370:7334/64 (host address is specified in the network address field, leaving no room for hosts)
2001:db8:85a3:0:0:8a2e:370:7334/120 (prefix length is less than 128, leaving room for hosts but not enough to specify the address itself)

Steps:

1. In cluster.yaml file add podSubnet value in under services as below and provide required "podSubnet" value
```
services:
  kubeadm:
    networking:
        podSubnet: x.x.x.x/x**
        serviceSubnet: y.y.y.y/y**
   kubernetesVersion: v1.25.2     
```

2. Run `kubemarine install --tasks="prepare.check"`

ER=AR: Success

3. Change **podSubnet** & **serviceSubnet** value to valid and invalid addresses and check the results.


alexarefev (commented):
What about the following case: the IP address notation is used instead of IP network notation. For instance: 192.168.0.1/16

nikhil1697:
> What about the following case: the IP address notation is used instead of IP network notation. For instance: 192.168.0.1/16

In this case the current code will raise a ValueError for the IPv4 subnet 192.168.0.1/16. This is because the ipaddress module expects the subnet to be in network address notation, where the subnet mask is represented by the number of bits set to 1 in the subnet mask.

alexarefev (approved):

PR #377, opened 2023-03-07
[PDSDNREQ-7070] HAproxy running with Root, should move to non-root

dmyar21:
### Description
* On the cluster, Haproxy runs from root, you need to make it run not from root, but from the haproxy user

Fixes # (issue)
PDSDNREQ-7070

### Solution
* Add new user and group parameters to `haproxy.cfg`

```
global
    log                 127.0.0.1 local2 debug
    user haproxy
    group haproxy
```
### How to apply
run
```
kubemarine install --tasks="deploy.loadbalancer.haproxy.configure"
```
or 

use patch `p2_fix_haproxy_user.py`

### Test Cases

**TestCase 1**

Test Configuration:

- Hardware: 
- OS: Ubuntu 20.04, 22.04, Centos 7
- Inventory: 

Steps:

1. 

Results:

| Before | After |
| ------ | ------ |
| Haproxy running with Root | Haproxy running with non-Root |

### Checklist
- [ ] Integration CI passed

### DIFF
--- a/kubemarine/templates/haproxy.cfg.j2
+++ b/kubemarine/templates/haproxy.cfg.j2
@@ -1,5 +1,7 @@
 global
     log                 127.0.0.1 local2 debug
+    user haproxy
+    group haproxy
 
 defaults
     log                 global
--- a/kubemarine/patches/__init__.py
+++ b/kubemarine/patches/__init__.py
@@ -23,8 +23,10 @@ from typing import List
 
 from kubemarine.core.patch import Patch
 from kubemarine.patches.p1_fix_nginx_ingress_k8s_1_25_x import FixNginxIngress
+from kubemarine.patches.p2_fix_haproxy_user import FixHaproxy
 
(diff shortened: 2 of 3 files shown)


kubemarine/templates/haproxy.cfg.j2
```diff
@@ -1,5 +1,7 @@
 global
     log                 127.0.0.1 local2 debug
+    user haproxy
```
koryaga, line 3: Pls make a patch

andrewluckyguy:
Corresponding user has been added
```
# grep user /etc/haproxy/haproxy.cfg
    user haproxy
```
but the command below returns empty value for haproxy.service
```
# systemctl show -pUser,UID haproxy.service
UID=[not set]
User=
```
As I know it means that service is running under root. 
Or maybe there is other way to find out which user is used.

ilia1243 (approved):

PR #345, opened 2023-01-25
Kubemarine: procedure for removing default taint on nodes having the roles of master and worker

dmyar21:
### Description

* It is necessary to remove the default taint on nodes that have the roles of master and worker at the same time. Instead of the old procedure that removed these taints in the `apply_taints` procedure

### Solution
* The procedure of deleting default taints on nodes having roles  of master and worker has been redesigned and transferred to another place (the moment of kubeadm init and join).
* Remove from `enrich_inventory` the procedure for adding standard taints for deletion in the `apply_taints` procedure.
* Update default.yaml
* Update documents
* Update Json schema
* Update unit test

### Note
* The possibility of updating to version 1.24.0 has been removed 
* When trying to upgrade to this version, an exception will be output

### Test Cases
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

TestCase 1

Test Configuration:

Hardware: Any
OS: Ubuntu 20.04
Inventory: MiniHA

Steps:

1. Run `install`.

Results:

| Before | After |
| ------ | ------ |
| Procedure of deleting taints in apply_taints  | Procedure of deleting taints in deploy_kubernetes_init  |

2. Run `upgrade`.

| Before | After |
| ------ | ------ |
| Procedure of deleting taints in apply_taints  | Procedure of deleting taints in deploy_kubernetes_init  |

### Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] Integration CI passed
- [ ] Unit tests. If Yes list of new/changed tests with brief description
- [ ] There is no merge conflicts

PR #397, opened 2023-03-28
[MANOPD-1] Update default version kubernetes

dmyar21:
### Description
* You need to update the default version of kubernetes in default.yaml

Fixes # (issue)
[MANOPD-86694]

### Solution
* Updated the default version of kubernetes to 1.26.3

### Checklist
- [ ] Integration CI passed

#### Unit tests


kubemarine/resources/configurations/defaults.yaml
```diff
@@ -717,7 +717,7 @@ rbac:
           annotations: {}
         type: kubernetes.io/service-account-token
 
-  admission: psp
+  admission: pss
```
ilia1243, line 720: This is incompatible change as Kubemarine suddenly starts to install different admission implementation on old clusters. I think that default admission implementation should be calculated dynamically based on kubernetes version if no explicit implementation is specified. @alexarefev

alexarefev: Definitely we need an improvement here

kubemarine/resources/configurations/defaults.yaml
```diff
@@ -42,7 +42,7 @@ services:
     # for patches functionality we need v1beta3 which is available since k8s v1.22 only
     apiVersion: '{% if "v1.21" in services["kubeadm"]["kubernetesVersion"] %}kubeadm.k8s.io/v1beta2{% else %}kubeadm.k8s.io/v1beta3{% endif %}'
     kind: ClusterConfiguration
-    kubernetesVersion: v1.24.11
+    kubernetesVersion: v1.26.3
```
ilia1243, line 45: Please also update schema/services/kubeadm.json. Also, I would like to remind that every update of defaults changes Kubemarine behaviour which can (highly unlikely) break some usages. https://github.com/Netcracker/KubeMarine/pull/345#discussion_r1100194109

ilia1243 (approved):

kubemarine/resources/configurations/defaults.yaml
dmyar21: update json schema

koryaga (approved):

PR #401, opened 2023-03-31
Manopd 85857 the restore procedure fails on the unpacking backup step

nikhil1697:
### Description
If resolve.conf is not specified in the cluster.yaml file then restore procedure was failing as chattr function is not supported on synbolic link
* 

### Solution
if we do not install resolv.conf, then backup & restore should leave resolv.conf on the nodes without changes so it should not take backup of resolv.conf and not restore it
* 

### How to apply
* [Install task/s](documentation/Installation.md#installation-tasks-description)
* Manual step 
* [Upgrade procedure](documentation/Maintenance.md#upgrade-procedure)

### Test Cases

**TestCase 1**

Test Configuration:

- Hardware: 
- OS: 
- Inventory: 

Steps:

1. 

Results:

| Before | After |
| ------ | ------ |
|  |  |

### Checklist
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/kubemarine/procedures/restore.py
+++ b/kubemarine/procedures/restore.py
@@ -20,6 +20,7 @@ import tarfile
 import time
 from collections import OrderedDict
 import yaml
+import shutil
 
 from kubemarine.core import utils, flow, defaults
 from kubemarine.core.action import Action
@@ -135,19 +136,25 @@ def restore_thirdparties(cluster):
 
     install.system_prepare_thirdparties(cluster)
 
-
+    
 def import_nodes(cluster):
     for node in cluster.nodes['all'].get_ordered_members_list(provide_node_configs=True):
-        node['connection'].put(os.path.join(cluster.context['backup_tmpdir'], 'nodes_data', '%s.tar.gz' % node['name']),
-                               '/tmp/kubemarine-backup.tar.gz')
-        cluster.log.debug('Backup \'%s\' uploaded' % node['name'])
+        node_backup_file = os.path.join(cluster.context['backup_tmpdir'], 'nodes_data', '%s.tar.gz' % node['name'])
+        if os.path.exists(node_backup_file):
+            node['connection'].put(node_backup_file, '/tmp/kubemarine-backup.tar.gz')
+            cluster.log.debug('Backup \'%s\' uploaded' % node['name'])
 
-    cluster.log.debug('Unpacking backup...')
-    result = cluster.nodes['all'].sudo(
-        'chattr -i /etc/resolv.conf; sudo tar xzvf /tmp/kubemarine-backup.tar.gz -C / --overwrite && sudo chattr +i /etc/resolv.conf')
-    cluster.log.debug(result)
+            if not cluster.procedure_inventory.get('backup_plan', {}).get('nodes', {}).get('resolv.conf'):
+                restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz --exclude=/etc/resolv.conf -C / --overwrite'
+            else:
+                restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz -C / --overwrite'
 
+            result = node['connection'].sudo(restore_command)
+            cluster.log.debug(result)
+        else:
+            cluster.log.debug('Backup file for node \'%s\' not found, skipping...' % node['name'])
 
+            
 def import_etcd(cluster: KubernetesCluster):
     etcd_all_certificates = cluster.procedure_inventory.get('restore_plan', {}).get('etcd', {}).get('certificates', {})
     etcd_cert = etcd_all_certificates.get('cert', cluster.globals['etcd']['default_arguments']['cert'])
--- a/kubemarine/procedures/backup.py
+++ b/kubemarine/procedures/backup.py
@@ -37,7 +37,6 @@ def get_default_backup_files_list(cluster: KubernetesCluster):
     keepalived_service = cluster.get_package_association('keepalived', 'service_name')
 
     backup_files_list = [
-        "/etc/resolv.conf",
         "/etc/hosts",
         "/etc/chrony.conf",
         "/etc/selinux/config",
@@ -54,6 +53,9 @@ def get_default_backup_files_list(cluster: KubernetesCluster):
         "/etc/systemd/system/kubelet.service"
     ]
 
+    if cluster.inventory["services"].get("resolv.conf") is not None:
+        backup_files_list.append("/etc/resolv.conf")
+
     cri_impl = cluster.inventory['services']['cri']['containerRuntime']
     if cri_impl == "docker":
         backup_files_list.append("/etc/docker/daemon.json")
@@ -62,6 +64,7 @@ def get_default_backup_files_list(cluster: KubernetesCluster):
         backup_files_list.append("/etc/crictl.yaml")
 
     return backup_files_list
+
 
 
 def prepare_backup_tmpdir(cluster):


alexarefev (approved):

heretic449 (approved):

kubemarine/procedures/restore.py
```diff
@@ -135,19 +136,25 @@ def restore_thirdparties(cluster):
...
-        cluster.log.debug('Backup \'%s\' uploaded' % node['name'])
-
-    cluster.log.debug('Unpacking backup...')
-    result = cluster.nodes['all'].sudo(
-        'chattr -i /etc/resolv.conf; sudo tar xzvf /tmp/kubemarine-backup.tar.gz -C / --overwrite && sudo chattr +i /etc/resolv.conf')
-    cluster.log.debug(result)
-
-
+        node_backup_file = os.path.join(cluster.context['backup_tmpdir'], 'nodes_data', '%s.tar.gz' % node['name'])
+        if os.path.exists(node_backup_file):
```
ilia1243, line 143: https://github.com/Netcracker/KubeMarine/blob/main/documentation/Maintenance.md#restore-procedure
> Everything should be the same, down to the names and addresses of the nodes, their amounts and roles

If the requirement is satisfied, the file `node_backup_file` always exists. Otherwise, we need to throw exception, and do no silently ignore the missed file.

kubemarine/procedures/restore.py
```diff
@@ -135,19 +136,25 @@ def restore_thirdparties(cluster):
...
+            node['connection'].put(node_backup_file, '/tmp/kubemarine-backup.tar.gz')
+            cluster.log.debug('Backup \'%s\' uploaded' % node['name'])
+
+            if not cluster.procedure_inventory.get('backup_plan', {}).get('nodes', {}).get('resolv.conf'):
```
ilia1243, line 147: procedure_inventory for `restore` procedure cannot have `backup_plan` property. See https://github.com/Netcracker/KubeMarine/blob/main/documentation/Maintenance.md#restore-procedure

ilia1243 (changes requested):

kubemarine/procedures/restore.py
nikhil1697: In that case, can we raise a FileNotFoundError exception with a message. 
 Like below 
 

>              result = node['connection'].sudo(restore_command)
>             cluster.log.debug(result)
>         else:
>             raise FileNotFoundError(f'Backup file for node \'{node["name"]}\' not found')

ilia1243: Ok, but it is a separate issue not related to your original PR. So please provide additional Test Case with
* *Result Before: something*
* *Result After: FileNotFoundError*

kubemarine/procedures/restore.py
nikhil1697: In that case, we can remove the following lines:

```
if not cluster.procedure_inventory.get('backup_plan', {}).get('nodes', {}).get('resolv.conf'):
                restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz --exclude=/etc/resolv.conf -C / --overwrite'
            else:
                restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz -C / --overwrite'
```

And replace them with just
`restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz -C / --overwrite'`



The entire code would look something like this, would this be OK?

```
def import_nodes(cluster):
    for node in cluster.nodes['all'].get_ordered_members_list(provide_node_configs=True):
        node_backup_file = os.path.join(cluster.context['backup_tmpdir'], 'nodes_data', '%s.tar.gz' % node['name'])
        if os.path.exists(node_backup_file):
            node['connection'].put(node_backup_file, '/tmp/kubemarine-backup.tar.gz')
            cluster.log.debug('Backup \'%s\' uploaded' % node['name'])

            if 'backup_plan' in cluster.procedure_inventory.get('restore', {}):
                raise ValueError('The procedure_inventory for restore procedure cannot have backup_plan property')

            restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz -C / --overwrite'
            if 'nodes' in cluster.procedure_inventory.get('restore', {}):
                nodes = cluster.procedure_inventory['restore']['nodes']
                if nodes.get('resolv.conf', False):
                    restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz --exclude=/etc/resolv.conf -C / --overwrite'

            result = node['connection'].sudo(restore_command)
            cluster.log.debug(result)
        else:
            raise FileNotFoundError('Backup file for node \'%s\' not found' % node['name'])
```

kubemarine/procedures/restore.py
```diff
@@ -135,19 +136,25 @@ def restore_thirdparties(cluster):
...
+                restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz --exclude=/etc/resolv.conf -C / --overwrite'
+            else:
+                restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz -C / --overwrite'
+
+            result = node['connection'].sudo(restore_command)
```
ilia1243, line 152: Previously we used nodes['all'] to unpack the archives on all nodes in parallel. We can still support parallel unpacking even in this case, but it is slightly more challenging. Consider using of `kubemarine.core.executor.RemoteExecutor`.
* It might also be possible to even upload the archive on nodes in parallel using the same RemoteExecutor instance.

kubemarine/procedures/restore.py
ilia1243: * There can never be cluster.procedure_inventory['restore']. All possible parameters are present in the mentioned documentation, or in JSON schema (see its mention in the same documentation).
* There previously was a `chattr -i /etc/resolv.conf` but it is removed in your last solution. Was it done intentionally?

kubemarine/procedures/restore.py
ilia1243: * get_ordered_members_list method is called exactly once before and after these changes, so the number of calls did not change.
* Using nodes['all'] to unpack archives on all nodes in each loop iteration is incorrect.
* RemoteExecutor is surely working feature, as it is very often used in Kubemarine in practice.

kubemarine/procedures/restore.py
ilia1243: Thank you, much better this time.

kubemarine/procedures/restore.py
```diff
@@ -135,19 +136,30 @@ def restore_thirdparties(cluster):
...
+    # Upload and restore backups for each node
+    for node in nodes:
+        node_backup_file = os.path.join(cluster.context['backup_tmpdir'], 'nodes_data', '%s.tar.gz' % node['name'])
+        if os.path.exists(node_backup_file):
+            node['connection'].put(node_backup_file, '/tmp/kubemarine-backup.tar.gz')
+            cluster.log.debug('Backup \'%s\' uploaded' % node['name'])
+
+            restore_command = 'sudo tar xzvf /tmp/kubemarine-backup.tar.gz -C / --overwrite'
+            
+            if 'resolv.conf' in tarfile.open(node_backup_file, 'r:gz').getnames():
```
ilia1243, line 154: The idea is good, but this does not work in case resolv.conf is present in the archive.

ilia1243 (changes requested):

kubemarine/procedures/restore.py
ilia1243: Example usage https://github.com/Netcracker/KubeMarine/blob/0.15.1/kubemarine/haproxy.py#L112

kubemarine/procedures/restore.py
nikhil1697: Okay To check if resolv.conf was extracted separately, i modified the code to search for the file in the tar archive and check if it was extracted to a different path than /etc/resolv.conf. Here's an updated implementation that does this:

ilia1243:
I looked at the problem a bit deeper and conclude that our current approach with restoration of resolv.conf is not correct. We need to restore dns configuration before working with internet resources, i. e. do it before `restore.thirdparties` task. Probably we can even change the order of tasks but this should be carefully investigated.

kubemarine/procedures/restore.py
```diff
@@ -135,19 +136,37 @@ def restore_thirdparties(cluster):
...
-
+        node_backup_file = os.path.join(cluster.context['backup_tmpdir'], 'nodes_data', '%s.tar.gz' % node['name'])
+        if os.path.exists(node_backup_file):
+            with tarfile.open(node_backup_file, 'r:gz') as tar:
+                # check if resolv.conf was extracted separately
+                resolv_conf_extracted = False
+                for member in tar.getmembers():
+                    if member.name.endswith('resolv.conf'):
+                        if member.path != '/etc/resolv.conf':
+                            resolv_conf_extracted = True
```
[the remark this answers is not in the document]
nikhil1697: These conditions are checking whether the resolv.conf file was extracted separately from the backup file or not.

The member.name.endswith('resolv.conf') condition checks whether the file name of the extracted file ends with the string "resolv.conf". If it does, then it is a resolv.conf file.

The member.path != '/etc/resolv.conf' condition checks whether the path of the extracted file is not equal to the standard location of the resolv.conf file on the node (/etc/resolv.conf). If the extracted resolv.conf file has a different path than the standard location, then it means that it was extracted separately.

If the resolv.conf file was extracted separately, it means that it was not restored as part of the backup extraction process, so it needs to be handled separately during the restore process. The code sets the resolv_conf_extracted flag to True to indicate that the resolv.conf file needs to be handled separately.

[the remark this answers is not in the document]
ilia1243: > the path of the extracted file is not equal to the standard location of the resolv.conf file on the node (/etc/resolv.conf)

Kubemarine does not support any other location except /etc/resolv.conf. The file should always be backuped from that location if necessary and restored to that location if necessary.

PR #407, opened 2023-04-04
[PSUPCLPL-11901] Add-validation-for-python

nikhil1697:
### Description

This improvement is a Python function that verifies if a specified method from a given module can be called with a specific set of arguments. It does so by attempting to import the module, checking if the method exists within it, and validating the method signature against the passed arguments. If the module cannot be imported or the method does not exist, a ValueError is raised. If the passed arguments do not match the method signature, a TypeError is raised.

### Solution
Improvement
* 

### How to apply
Code change

### Test Cases

1. Configure any plugin with plugin installation step with python method.
 for example 
```
      procedures:
        - python:
            module: plugins/nginx_ingress.py
            method: manage_custom_certificate
        - python:
            module: plugins/builtin.py
            method: apply_yaml
            arguments: 
              plugin_name: nginx-ingress-controller
```
2. Run kubemarine install --without-act and check that it succeeds
3. Change python method
   for example change name of the method which is not available but not the plugin configuration
```
           - python:
            module: plugins/builtin.py
            method: apply_yaml_test
            arguments: 
              plugin_name: nginx-ingress-controller
```
4. Run kubemarine install --without-act and check that it fails

### Checklist
- [ ] Integration CI passed

#### Unit tests


alexarefev (approved):

kubemarine/plugins/__init__.py
```diff
@@ -522,8 +523,33 @@ def apply_expect(cluster, config, plugin_name=None):
...
+        spec.loader.exec_module(module)
+    except Exception as e:
+        raise ValueError(f"Could not import module {module_path}: {e}")
+
+    # Check if the method exists
+    if not hasattr(module, method_name):
+        raise ValueError(f"Module {module_path} does not have method {method_name}")
+
+    # Get the method object
+    method = getattr(module, method_name)
```
ilia1243, line 543: There is a code duplication. Instead, you could add new method `get_python_method_args() -> Tuple[callable, dict]` that will 1) import module and perform all validations; 2) return method and args. Use the method in both verify_python and apply_python.

kubemarine/plugins/__init__.py
```diff
@@ -522,8 +523,33 @@ def apply_expect(cluster, config, plugin_name=None):
...
-    return
+    module_path, _ = utils.determine_resource_absolute_file(step['module'])
+    method_name = step['method']
+    method_arguments = step.get('arguments', {})
+
+    # Try to import the module
+    try:
+        spec = importlib.util.spec_from_file_location('module', module_path)
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
```
ilia1243, line 534: Now the module is executed twice that might be not expected. I suggest to ensure to execute the module only once.

ilia1243: This is still executed twice, when called from apply_python and verify_python. I understand that we already do that for our own modules https://github.com/Netcracker/KubeMarine/blob/0.18.0/kubemarine/resources/configurations/defaults.yaml#L542-L547, but this is becoming worse for custom modules if they are used only once. If it is hard to fix, we can introduce restriction to documentation that execution should be idempotent, duplicates of static variables and classes might appear and so on.

kubemarine/plugins/__init__.py
```diff
@@ -573,22 +574,47 @@ def apply_expect(cluster, config, plugin_name=None):
...
+
+    return method, method_arguments
+
+
 def verify_python(cluster, step):
-    # TODO: verify fields types and contents
-    return
+    module_path, _ = utils.determine_resource_absolute_file(step['module'])
+    method_name = step['method']
+    method_arguments = step.get('arguments', {})
```
ilia1243, line 602: This could also moved to get_python_method_args

kubemarine/plugins/__init__.py
nikhil1697: I tried skipping exec_module for verify python, but it was giving error method not found so without getting a method reference it will fail to perform validation of method signature.

PR #411, opened 2023-04-05
MANOPD-87063 Configure values_file and values simultaneously for helm plugins

ilia1243:
### Description
* It should be possible to configure `values_file` and `values` simultaneously for helm plugins.
* Both sources of values should be applied, and the values from `values` should have higher priority.

### Solution
* Changes are implemented in `plugins.apply_helm` according to the requriements.
* Added patch which removes probably unused `values_file` section from old inventories. Now this parameter is always used, if specified.

### How to apply
Run `migrate_kubemarine --force-apply helm_values`

### Test Cases

**TestCase 1**

Steps:

1. Configure both `values_file` and `values` sections for helm plugin.
2. Run `install` procedure.

Results:

| Before | After |
| ------ | ------ |
| Only `values` parameter is used | Both parameters are used, and `values` has higher priority. |

**TestCase 1**

Steps:

1. Configure both `values_file` and `values` sections for helm plugin.
2. Run `install --tasks deploy.plugins` using older Kubemarine version.
    ER: Only `values` parameter is used.
3. Run `migrate_kubemarine --force-apply helm_values` using new Kubemarine.
    ER: `values_file` parameter is removed from inventory.

### Checklist
- [x] Integration CI passed

#### Unit tests
Added test_helm.py which checks processing of values.yaml.


test/unit/plugins/test_helm.py
```diff
@@ -0,0 +1,64 @@
...
+import tempfile
+import unittest
+
+import yaml
+
+from kubemarine import demo, plugins
+
+
+class TestHelm(unittest.TestCase):
+    def test_process_chart_values_override_precedence(self):
```
Imadzuma, line 25: Could you add more tests for only `values` and `values_file` defined?

ilia1243: Done

Imadzuma (approved):

koryaga (approved):

PR #424, opened 2023-04-21
[MANOPD-84744] fix remove_node.py in case of custom etc_hosts value

n549:
### Description
**I.** In case of custom-defined `etc_hosts` section in cluster.yaml `kubemarine remove_node` fails with an error:
```
  File "/home/nat/platform/KubeMarine/kubemarine/coredns.py", line 31, in enrich_add_hosts_config
    if not inventory['services']['coredns']['configmap'].get('Hosts'):
  File "/usr/local/lib/python3.9/site-packages/ruamel/yaml/comments.py", line 927, in __getitem__
    return ordereddict.__getitem__(self, key)
KeyError: 'coredns'
```
**II.** By default we put cluster nodes ip addresses and names into `coredns.Hosts` configmap, but in case of `services.coredns.configmap.Hosts` is defined in `cluster.yaml`, it doesn't happened.

Fixes # (issue)
MANOPD-84744

### Solution
**I.** RC of the issue is that `coredns.enrich_add_hosts_config()` tries to [call](https://github.com/Netcracker/KubeMarine/blob/d02cfced2f0da2ee9a5a99fbd11f53c610a257df/kubemarine/coredns.py#L24) `inventory['services']['coredns']['configmap']` object which is absent in the inventory which is used when this function is called from [remove_node.py](https://github.com/Netcracker/KubeMarine/blob/d02cfced2f0da2ee9a5a99fbd11f53c610a257df/kubemarine/procedures/remove_node.py#L122)

So `coredns.enrich_add_hosts_config()` is removed from `removed_node.remove_node_finalize_inventory()`

**II.** Work with `etc_hosts` and `coredns.configmap.Hosts` is reworked.
- new inventory parameter `services.etc_hosts_generated` is introduced. It's filled by Kubemarine with cluster nodes ip addresses and names. It's present in `cluster_finalized.yaml` but only for informational purpose and it's reset during kubemarine work.
- `services.etc_hosts` contains only default data from `defaults.yaml` and data defined by user in `cluster.yaml`;
- new parameter `etc_hosts_part` is added to `system.generate_etc_hosts_config()` is added, it defines from where (`etc_hosts` or `etc_hosts_generated`) data should be taken to add lines to `/etc/hosts`;
- `install.system_prepare_dns_etc_hosts() creates `/etc/hosts` at the nodes which consists of 2 parts: one is generated from `services.etc_hosts` and another is generated from `services.etc_hosts_generated`;
- `remove_node_finalize_inventory()` doesn't remove info about a deleting node from `services.etc_hosts` if it's present there. A deleting node is removed from `services.etc_hosts_generated` though;
- new inventory parameter `services.coredns.add_etc_hosts_generated` is introduced. It's boolean and defines whether we should add information about cluster nodes' ip addresses and names into `coredns.Hosts` configmap. Default is `true`;
- `coredns.enrich_add_hosts_config()` function is deleted. It's not needed any more because `services.coredns.add_etc_hosts_generated` content is added to `coredns.configmap.Hosts` during the configmap generation;
- added default value for `coredns.configmap.Hosts` containing localhost records;
- coredns unit tests are updated according to the new logic;
- Installation.md is updated respectively.

### How to apply
Run `remove_node` with cluster.yaml containing `etc_hosts` section.

### Test Cases
**I.**
1. Run `remove_node` procedure with cluster.yaml containing `etc_hosts` section with custom data.
ER: Procedure finished successfully.

**II.**
1. Deploy cluster with custom `services.etc_hosts`.
ER: Deployment is successfull, /etc/hosts and the nodes contains custom and default data from `services.etc_hosts` as well as records about cluster nodes. Configmap `coredns` contains `Hosts` part with data about cluster nodes.

2. Remove a node from the cluster. Information about this node is absent in custom `services.etc_hosts`.
ER: The node is removed successfully, its data is removed from /etc/hosts at the nodes and from `coredns` configmap.

3. Run task `prepare.dns.etc_hosts` with `cluster.yaml` containing some custom data in `services.coredns.configmap.Hosts`.
ER: Configmap `coredns` contains the custom data and data about cluster nodes.

4. Set `services.coredns.add_etc_hosts_generated` to `false`, remove `services.coredns.configmap.Hosts` from `cluser.yaml` and run `prepare.dns.etc_hosts` task. 
ER: `coredns` configmap contains empty `Hosts` section.

### Checklist
- [x] Integration CI passed

#### Unit tests
Adjusted tests in test_coreds.py: `test_add_hosts_config`, `test_configmap_generation`, `test_configmap_generation_with_hosts`, test `test_already_defined_hosts_config` is renamed to `test_already_defined_hosts_config_and_not_add_etc_hosts_generated` and also adjusted to reflect new changes.


test/unit/test_coredns.py
```diff
@@ -24,7 +24,7 @@ class CorednsDefaultsEnrichment(unittest.TestCase):
     def test_add_hosts_config(self):
         inventory = demo.generate_inventory(**demo.MINIHA_KEEPALIVED)
         cluster = demo.new_cluster(inventory)
-        generated_hosts = system.generate_etc_hosts_config(cluster.inventory)
+        generated_hosts = system.generate_etc_hosts_generated_config(cluster.inventory)
```
ilia1243, line 27: Could you also add more tests to cover new behaviour?

kubemarine/coredns.py
```diff
@@ -21,8 +21,16 @@
 
 
 def enrich_add_hosts_config(inventory, cluster):
-    if not inventory['services']['coredns']['configmap'].get('Hosts'):
-        inventory['services']['coredns']['configmap']['Hosts'] = system.generate_etc_hosts_config(inventory, cluster)
+    if inventory['services']['coredns']['add_etc_hosts_generated']:
+        if not inventory['services']['coredns']['configmap'].get('Hosts'):
+            inventory['services']['coredns']['configmap']['Hosts'] = system.generate_etc_hosts_generated_config(cluster.inventory, cluster)
+        else:
+            inventory['services']['coredns']['configmap']['Hosts'] += system.generate_etc_hosts_generated_config(cluster.inventory, cluster)
```
ilia1243, line 28: We stopped adding `localhost` to the `services.coredns.configmap.Hosts`. Was it intended?

kubemarine/resources/schemas/definitions/services.json
```diff
@@ -54,6 +54,9 @@
     "etc_hosts": {
       "$ref": "services/etc_hosts.json"
     },
+    "etc_hosts_generated": {
```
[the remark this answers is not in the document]
n549: It can be redefined but it makes no sense because `services.etc_hosts_generated` is set empty just before filling with nodes information. It can be used for a quick reference in cluster_finalized.yaml if necessary, though.
Final `etc_hosts` and `coredns_configmap.yaml` are dumped into ./dump/ folder and can be also revised there.

kubemarine/system.py
```diff
@@ -48,30 +48,33 @@ def verify_inventory(inventory, cluster):
 
 
 def enrich_etc_hosts(inventory, cluster):
+# enrich only etc_hosts_generated object, etc_hosts remains as it is
+
     control_plain = inventory['control_plain']['internal']
 
+    # take custom name of control_plain from etc_hosts, if any
     control_plain_names = inventory['services']['etc_hosts'].get(control_plain, [])
```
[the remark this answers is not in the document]
n549: Looks like we should take custom name for control_plane from `services.etc_hosts` if it's present there. If there is none, we don't take it and proceed with default set of node names.

kubemarine/coredns.py
n549: Actually `localhost` was added to `services.coredns.configmap.Hosts`  only when no custom `services.coredns.configmap.Hosts` were present. 
Now I've added `localhost` to default settings of `services.coredns.configmap.Hosts` so it's added to the configmap in case of a user doesn't override this default setting.

kubemarine/system.py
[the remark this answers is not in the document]
ilia1243: If it is present there, it will be added to /etc/hosts by system.generate_etc_hosts_config(). It will not be added to `coredns.configmap.Hosts`, but we decided to make them independent.

kubemarine/system.py
```diff
@@ -48,30 +48,36 @@ def verify_inventory(inventory, cluster):
...
         if 'remove_node' in node['roles']:
             continue
 
-        internal_node_ip_names = inventory['services']['etc_hosts'].get(node['internal_address'], [])
+        internal_node_ip_names = inventory['services']['etc_hosts_generated'].get(node['internal_address'], [])
         internal_node_ip_names.append("%s.%s" % (node['name'], cluster.inventory['cluster_name']))
         internal_node_ip_names.append(node['name'])
         internal_node_ip_names = list(OrderedSet(internal_node_ip_names))
-        inventory['services']['etc_hosts'][node['internal_address']] = internal_node_ip_names
+        inventory['services']['etc_hosts_generated'][node['internal_address']] = internal_node_ip_names
```
ilia1243, line 73: This code becomes too complex. It was designed to add new names to list of custom names in `etc_hosts`, but we do not support custom names in `etc_hosts_generated`.

kubemarine/system.py
```diff
@@ -299,11 +330,11 @@ def generate_etc_hosts_config(inventory, cluster=None):
 
     ignore_ips = list(set(ignore_ips))
 
-    for ip in list(inventory['services']['etc_hosts'].keys()):
+    for ip in list(inventory['services']['etc_hosts_generated'].keys()):
         if len(ip) > max_len_ip:
             max_len_ip = len(ip)
 
-    for ip, names in inventory['services']['etc_hosts'].items():
+    for ip, names in inventory['services']['etc_hosts_generated'].items():
```
ilia1243, line 337: We no longer need to collect and use ignore_ips collection as etc_hosts_generated cannot have IPs that are going to be removed. In fact, we can have the only method with parameter to generate both configs ` generate_etc_hosts_config('etc_hosts' | 'etc_hosts_generated')`

kubemarine/resources/configurations/defaults.yaml
```diff
@@ -341,6 +344,9 @@ services:
           forward:
             - .
             - /etc/resolv.conf
+      Hosts: '127.0.0.1 localhost localhost.localdomain
+        {% if not nodes[0]["internal_address"]|isipv4 %}::1 localhost localhost.localdomain{% endif %}
```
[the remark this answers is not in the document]
n549: fixed.

kubemarine/resources/schemas/definitions/services/coredns.json
```diff
@@ -3,6 +3,11 @@
   "type": "object",
   "description": "Configure the Coredns service and its DNS rules in the Kubernetes cluster",
   "properties": {
+    "add_etc_hosts_generated": {
+      "type": "boolean",
+      "default": true,
+      "description": "Enables to add autogenerated records about cluster nodes to Hosts part of coredns confifmap"
```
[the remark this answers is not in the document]
n549: thanks, fixed.

kubemarine/system.py
n549: reworked.

ilia1243 (approved):

PR #434, opened 2023-05-04
[MANOPD-87147] Improving colors

akor0421:
### Description
TestCase and TestSuite classes from [testsuite.py](https://github.com/Netcracker/KubeMarine/blob/main/kubemarine/testsuite.py) always use colors to display text and background, even if `colorize=false` is specified in the arguments.

### Solution
Now, when calling TestCase, execution arguments are also passed, after which `colorize=false` is checked  in them for stdout.

Also now in Windows, colors are also used by default. And the small doc update about using Windows Terminal in Windows for correct color display.

### Test Cases
Run check_paas/check_iaas procedure with `colorize=false` for stdout
` kubemarine check_paas --log="stdout;level=warning;colorize=false"`

### Checklist
- [ ] Integration CI passed

### DIFF
--- a/kubemarine/procedures/check_iaas.py
+++ b/kubemarine/procedures/check_iaas.py
@@ -36,7 +36,7 @@ from kubemarine.testsuite import TestSuite, TestCase, TestFailure, TestWarn
 from kubemarine.core.group import NodeGroupResult
 
 def connection_ssh_connectivity(cluster):
-    with TestCase(cluster.context['testsuite'], cluster.context['execution_arguments'], '001', 'SSH', 'Connectivity', default_results='Connected'):
+    with TestCase(cluster, '001', 'SSH', 'Connectivity', default_results='Connected'):
         failed_nodes = []
         for node in cluster.nodes['all'].get_ordered_members_list(provide_node_configs=True):
             try:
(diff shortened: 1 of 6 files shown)


n549 (approved):

koryaga:
@OleksandrIaresko  pls validate

kubemarine/procedures/check_iaas.py
```diff
@@ -36,7 +36,7 @@
 from kubemarine.core.group import NodeGroupResult
 
 def connection_ssh_connectivity(cluster):
-    with TestCase(cluster.context['testsuite'], '001', 'SSH', 'Connectivity', default_results='Connected'):
+    with TestCase(cluster.context['testsuite'], cluster.context['execution_arguments'], '001', 'SSH', 'Connectivity', default_results='Connected'):
```
ilia1243, line 39: Maybe we could pass the only property cluster.context to TestCase constructor, and fetch both properties `testsuite`, 'execution_arguments' from it there.

kubemarine/testsuite.py
```diff
@@ -138,6 +154,18 @@ def get_summary(self, show_description=False, show_hint=False, show_minimal=Fals
...
+        if self.args.get('log') is not None:
+            for argsgroup in self.args.get('log'):
+                if argsgroup[0].split(";")[0] != 'stdout':
+                    continue
+                for argument in argsgroup[0].split(";"):
+                    if argument == 'colorize=false':
+                        return False
+            return True
+        else:
+            return True
```
ilia1243, line 167: Alternatively, more simpler:
```suggestion
    def check_color(self):
        for handler in cluster.log.handlers:
            if isinstance(handler, log.StdoutHandler) and handler.formatter.colorize:
                return True
        return False
```
provided that either cluster, or cluster.log is passed to TestCase instance.

kubemarine/procedures/check_iaas.py
akor0421: Done, passed `cluster` to TestCase to take `cluster.log` too

ilia1243 (approved):

Issue #269, opened 2022-10-27
Implement procedure to switch haproxy mode


ilia1243:
It is necessary to implement maintenance procedure which switches haproxy mode. See [maintenance mode](https://github.com/Netcracker/KubeMarine/blob/0.8.0/documentation/Installation.md#maintenance-mode) and [maintenance type](https://github.com/Netcracker/KubeMarine/blob/0.8.0/documentation/Installation.md#maintenance-type).

Also, during cluster installation over an existing cluster, the mode should be reverted to default (not maintenance).

pranavcracker:
Maintenace mode support for haproxy load balancer is already present in current code.

With _maintenance_mode: True_ in haproxy properties in services and _maintenance-type: "not bind"_ paramter in vrrp properties, we can set the haproxy in maintenance mode.

We need a mechanism for setting the maintenace mode on/off on haproxy.

pranavcracker:
We have a draft plan to create a new procedure named as "_haproxy_mntc.py_" and working on creating this procedure which will *enable/disable* maintenance mode on HAProxy.

This procedure will do the following-
 1. Check if maintenance mode is supported by the cluster be checking presence of below parameter
```
loadbalancer:
    haproxy:
      maintenance_mode: True
```
If above mentioned parameter is missing, exit the procudure with error msg.

 2.  Do the config chnages required to enable/disable maintenance mode 
 3.  Perform daemon reload and service restart for haproxy

Usage of this new procedure will be like 
```
kubemarine haproxy_mntc enable/disable

PR #435, opened 2023-05-04
[PSUPCLPL-12244] HAProxy Maintenance Mode

pranavcracker:
### Description
Adding a new procedure to Kubemarine distribution for managing the maintenance mode for HAProxy load balancers.

Fixes # (issue)
https://github.com/Netcracker/KubeMarine/issues/269 

### Solution

* We are developing a new procedure which will help us to enable/disable maintenance mode by a single command as below

### How to apply
Once the changes mentioned in this PR, we will have a automated procedure to enable or disable the maintenance mode on HAProxy load balancers. 
* Go to DVM --> Maintenance Job --> Build with parameters --> Select "haproxy_mntc" procedure from the drop down list --> provide the required argument (--mode enable/disable) --> Run the job
* Manual step -
Login to CLI and navigate to kubemarine directory --> run the following command
```
kubemarine haproxy_mntc --mode enable/disable
```

### Test Cases

**TestCase 1**

Run **haproxy_mntc** procedure for the cluster where maintenance mode is not supported.
`kubemarine haproxy_mntc --mode enable`
ER- procedure should fail with error message "Skiping haproxy maintenance procedure as cluster is not supported for HAProxy Maintenance mode."

**TestCase 2**

Run **haproxy_mntc** procedure for the cluster where maintenance mode is supported.

1. `kubemarine haproxy_mntc --mode enable`
2. `kubemarine haproxy_mntc --mode disable`

ER- Procedure should succeed  with enabling or disabling the maintenance mode on haproxy load balancer.

### Checklist
- [x] Integration CI passed


kubemarine/procedures/haproxy_mntc.py
```diff
@@ -0,0 +1,111 @@
...
+
+import os
+from collections import OrderedDict
+import time
+from kubemarine.core import flow
+from kubemarine.core.action import Action
+from kubemarine.core.resources import DynamicResources
+from kubemarine.haproxy import is_maintenance_mode
+
+def haproxy_mntc(cluster):
```
ilia1243, line 23: I suggest another approach, which is more complex but more Kubemarine-like in my opinion.
1. Introduce new inventory
   ```yaml
   services:
     loadbalancer:
       haproxy:
         maintenance_mode:
           install: False
           enabled: False
           config_location: /etc/haproxy/haproxy_mntc.cfg
           keep_updated: True
   ```
   The above are default values (defaults.yaml)
   install - whether to install maintenance config haproxy_mntc.cfg. Has the same meaning as the `services.loadbalancer.haproxy.maintenance_mode` currently.
   enabled - whether to enable the maintenance mode on balancer nodes.
   config_location - moved from `services.loadbalancer.haproxy.mntc_config_location`
   keep_updated - If false, `kubemarine install --tasks deploy.loadbalancer.haproxy.configure` will not switch mode (see p. 2 below).
2. `install --tasks deploy.loadbalancer.haproxy.configure` configures and installs both configs if `services.loadbalancer.haproxy.keep_configs_updated == True` and switches mode if `maintenance_mode.keep_updated == True`
3. add/remove procedures only configure and install both configs if `services.loadbalancer.haproxy.keep_configs_updated == True`
4. New `haproxy_mntc` is not necessary in this approach, but if implemented, it only switches mode ignoring value of `maintenance_mode.keep_updated`. It has procedure inventory of the similar format as in p. 1 but supporting only `maintenance_mode.enabled`.
5. Write patch (see [Kubemarine Migration Procedure](https://github.com/Netcracker/KubeMarine/blob/main/documentation/Maintenance.md#kubemarine-migration-procedure) that changes format of `haproxy.maintenance_mode` in inventory, and sets `maintenance_mode.keep_updated=False` for backward compatibility.

pranavcracker: Hi,
Yes, the suggested approach is more complex.
But the approach used in this PR is simpler and more effective as well.
We are able to enable/disable maintenance mode on haproxy with ease. 
If maintenance mode is not supported by the cluster this procedure will not work in that cluster and will give the appropriate error message.

koryaga: Let's have an arch discussion about this topic. Until then - postponed.

koryaga: Decided to cancel current implementation in the favor of provided above proposal. The further implementation should be start with detailed ADR and based on this comment

koryaga:
decided to redesign

PR #446, opened 2023-05-22
[MANOPD-88530] Update task "services.system.modprobe.rules" for different os in Check_PaaS 

dmyar21:
### Description
* You need to add an exception to check modprobe in check paas for a cluster with different operating systems

Fixes # (issue)
[MANOPD-88530]

### Solution
* Added an exception to check modprobe on a cluster from with different operating systems

### Checklist
- [ ] Integration CI passed

#### Unit tests


kubemarine/system.py
```diff
@@ -610,11 +610,15 @@ def is_modprobe_valid(group):
...
-                is_valid = False
+    if os_family != 'multiple':
+        for module_name in group.cluster.inventory['services']['modprobe'][os_family]:
+            for conn, result in verify_results.items():
+                if module_name not in result.stdout:
+                    log.debug('Kernel module %s not found at %s' % (module_name, conn.host))
+                    is_valid = False
+    else:
+        log.debug('Several different operating systems have been found')
+        is_valid = False
```
ilia1243, line 621: It is potentially possible to check `is_modprobe_valid` being on multiple OS family. If we do not want to support this, we can prohibit any operation with modprobe explicitly inside
* check_paas.verify_modprobe_rules. The similar way as it was done in [check_paas.system_packages_versions](https://github.com/Netcracker/KubeMarine/blob/0.17.0/kubemarine/procedures/check_paas.py#L166)
* system.setup_modprobe. The similar restriction was for [audit.install](https://github.com/Netcracker/KubeMarine/blob/0.11.0/kubemarine/audit.py#L39) in past, but currently is eliminated.

dmyar21: it was decided to skip the modprobe check

ilia1243 (approved):

PR #457, opened 2023-06-06
Edit helm command to install chart in apply_helm function

krang404:
### Description
If using custom procedure with helm installation in new namespace of k8s cluster, you always get "CRITICAL KME0001: Unexpected exception". It happens because "`helm install`" calls  without flag `--create-namespace`

### Solution
In **def apply_helm** kubemarine/plugins/__init__.py, line 788, add flag `--create-namespace`. It will not breaking procedure, because helm handle exception if namespace already exist.   

### Checklist
- [ ] Integration CI passed


andrewluckyguy (approved):
Verified by QA in scenario with helm plugin installation, when shell command 'kubectl create namespace' wasn't executed on purpose

koryaga (approved):

PR #463, opened 2023-06-21
[MANOPD-89188] Type annotations and checks by mypy

### LINKED ISSUE #142: Get rid of cluster.inventory["nodes"][i]["connection"]
We currently store fabric connection object directly in inventory. See defaults.apply_defaults():

```
cluster.inventory["nodes"][i]["connection"] = cluster.make_group([address])
```

This already caused some problems:

- necessity to remove this object in finalize inventory procedure (defauls.prepare_for_dump())
- after inventory is finalized, methods which require this key, become not working.
- if defauls.prepare_for_dump() is invoked with copy=True, this fabric connection is deep copied, which is undefined behaviour (at least, untill we prove that fabric allows such copying and this will not be broken during future upgrades of fabric).

Instead, I suggest to either entirely get rid of such caches of connections, or use other storage, for example cluster.context.

ilia1243:
### Description
* Turn on type checks by mypy.

Fixes #142

### Solution
* Added many type annotations. Added running of `mypy` in CI. 
* Remove `provide_node_configs` option for NodeGroup's group methods. Split into two `get_ordered_members_configs_list` and `get_ordered_members_list` methods with different return types.
* Introduce abstract and generic `AbstractGroup` that holds run/sudo/get/put connection methods, and their common implementations, and all methods to operate with groups.
    * `NodeGroup` implementation executes connection methods immediately and return the result. It is also able to work around common errors, reboot nodes.
    * `DeferredGroup` impementation queues remote commands for connection methods and executes them during flush of RemoteExecutor.
* `NodeGroupResult` is made Generic and two implementations are introduced holding generic and simple runners result.
* Introduced custom `RunnersResult` and `GroupException` classes, removed dependency on fabric GroupResult.
* Support async transferring of files and get rid of time.sleep(). https://github.com/Netcracker/KubeMarine/commit/815df335fd697485ac0d3a63f8eaee23c792979c
* RemoteExecutor now stops execution for connections if errors occur.
* Get rid of `node['connection']` key for better type checks.
* fabric Connection instances are no longer used as dictionary keys.
* fabric Connection is now used only in executor.py and connections.py.
* Resolve cyclic dependencies of RemoteExecutor and NodeGroup.
* All NodeGroup connection methods now accept explicit subset of supported kwargs.
* Remove `is_async` arg from `run` / `sudo`. Now all execution is async. To run sync commands, it is necessary to explicitly iterate over nodes controlling the order.
* Remove `RemoteExecutor.get_last_results_str`. Use `RemoteExecutor.get_merged_runners_result()` with tokens instead.
* Fix enrichment and finalization of `add_node` procedure. https://github.com/Netcracker/KubeMarine/commit/c0ffdcda50e349370ac14fba724598c39a206fbe
* Other major rework and minor bugfixes in RemoteExecutor and NodeGroup.

### Other notes to pay attention to
* Link to mypy documentation https://mypy.readthedocs.io/en/latest/index.html
    * [Using mypy with an existing codebase](https://mypy.readthedocs.io/en/latest/existing_code.html)
    * Also some complex type annotation features were used, [Generics](https://mypy.readthedocs.io/en/latest/generics.html), [Protocols and structural subtyping](https://mypy.readthedocs.io/en/latest/protocols.html)
* More strict mypy check were enabled globally, see `[tool.mypy]` section of pyproject.yaml.
* Even stricter mypy checks were enabled for many modules. Typically, it was done for modules that are completely type annotated. See `[[tool.mypy.overrides]]` sections of pyproject.yaml. This prevents unintended adding of new not annotated methods.

### Regresion Tests
Run all procedures to test the regression.

### Checklist
- [x] Integration CI passed

#### Unit tests
test_group.py, test_executor.py - add new tests for async put.


kubemarine/core/executor.py
```diff
@@ -118,288 +133,206 @@ def reparse_results(self, results, batch):
...
+                futures: Dict[str, concurrent.futures.Future] = {}
 
-                def safe_exec(result_map: Dict[str, Any], key: str, call: Callable[[], Any]):
+                def safe_exec(result_map: Dict[str, Any], host: str, call: Callable[[], Any]) -> None:
                     try:
-                        # sleep required to avoid thread starvation
-                        time.sleep(0.1)
-                        result_map[key] = call()
-                        time.sleep(0.1)
+                        result_map[host] = call()
```
ilia1243, line 292: time.sleep was an unreliable workaround to fix bug in async transferring of files. The bug is fixed in https://github.com/Netcracker/KubeMarine/commit/815df335fd697485ac0d3a63f8eaee23c792979c

kubemarine/resources/configurations/globals.yaml
```diff
@@ -48,7 +48,7 @@ nodes:
       timeout: 10
       retries: 30
   boot:
-    reboot_command: 'reboot 2>/dev/null >/dev/null &'
+    reboot_command: 'systemctl stop sshd && sudo reboot 2>/dev/null >/dev/null'
```
ilia1243, line 51: Sometimes reboot is stuck. The problem is relatively stably reproduced on some deployer environments, but is still floating.

Just after reboot in [system.perform_group_reboot](https://github.com/Netcracker/KubeMarine/blob/0.18.2/kubemarine/system.py#L377) we [disconnect](https://github.com/Netcracker/KubeMarine/blob/0.18.2/kubemarine/core/group.py#L687) fabric Connections and then try to run `last reboot`. Running of commands internally creates paramiko client via [Connection.open()](https://github.com/fabric/fabric/blob/2.6.0/fabric/connection.py#L636) and then `Connection.run` opens paramiko channel.

For unknown reason in some cases, paramiko client is opened successfully just after reboot (probably because reboot procedure did not kill sshd daemon yet), but [opening of channel](https://github.com/paramiko/paramiko/blob/2.11.1/paramiko/transport.py#L1045-L1055) is stuck for 1 hour.

```
Thread 0x00007f88237fe700 (most recent call first):
  File "/usr/local/lib/python3.11/threading.py", line 324 in wait
  File "/usr/local/lib/python3.11/threading.py", line 622 in wait
  File "/usr/local/lib/python3.11/site-packages/paramiko/transport.py", line 1046 in open_channel
  File "/usr/local/lib/python3.11/site-packages/paramiko/transport.py", line 920 in open_session
  File "/usr/local/lib/python3.11/site-packages/fabric/connection.py", line 700 in create_session
  File "/usr/local/lib/python3.11/site-packages/fabric/connection.py", line 30 in opens
  File "<decorator-gen-2>", line 2 in create_session
  File "/usr/local/lib/python3.11/site-packages/fabric/runners.py", line 36 in start
  File "/usr/local/lib/python3.11/site-packages/invoke/runners.py", line 431 in _run_body
  File "/usr/local/lib/python3.11/site-packages/invoke/runners.py", line 380 in run
  File "/usr/local/lib/python3.11/site-packages/invoke/context.py", line 102 in _run
  File "/usr/local/lib/python3.11/site-packages/fabric/connection.py", line 723 in run
  File "/usr/local/lib/python3.11/site-packages/fabric/connection.py", line 30 in opens
  File "<decorator-gen-3>", line 2 in run
  File "/usr/local/lib/python3.11/concurrent/futures/thread.py", line 58 in run
  File "/usr/local/lib/python3.11/concurrent/futures/thread.py", line 83 in _worker
  File "/usr/local/lib/python3.11/threading.py", line 975 in run
  File "/usr/local/lib/python3.11/threading.py", line 1038 in _bootstrap_inner
  File "/usr/local/lib/python3.11/threading.py", line 995 in _bootstrap
```

I suggest to stop the sshd daemon just before the reboot. In practice, Connection.open() was never successful in many attempts, but it uses the connection timeout of 10 seconds that we pass. It is thus finished after a number of attempts.

Also removed `&` because running of `reboot` in background does not work. https://www.fabfile.org/faq.html#why-can-t-i-run-programs-in-the-background-with-it-makes-fabric-hang

This has the drawback that if sshd dameon is stopped but reboot does not happen, connection via ssh will be lost.

heretic449 (approved):

koryaga (approved):

PR #464, opened 2023-06-26
Implement dry-run for install/upgrade process

Deo121996:
### Description
This improvement is to enable dry-run for installation and upgrade procedure of Kubemarine using the existing --without-act option
* This dry-run implementation will perform all the possible checks and generate all the possible files in the dumps directory to simulate the actual install/upgrade process.
* It will not make changes to actual nodes but perform all the possible checks by executing read commands.
* User can simulate the actual procedure by running a dry run before executing the actual procedure.

### Solution
Improvement
* 

### How to apply
Code change

### Test Cases

**TestCase 1**
Steps:
1. Configure a valid cluster.yaml for the all-in-one cluster.
2. Run kubemarine install --without-act and check if the process succeeds and configuration files are created in the dumps directory.

**TestCase 2**
Steps:
1. Pass invalid plugin  yaml template inside kubemarine/plugins/yaml directory.
2. Run kubemarine install --without-act and check if the process fails and configuration files are created in the dumps directory.

**TestCase 3**
Steps:
1. Configure a valid procedure.yaml for the upgrade.
2. Run kubemarine upgrade procedure.yaml --without-act and check if the process succeeds and configuration files are created in the dumps directory.

### Checklist
- [ ] Integration CI passed


Deo121996:
**Results**

test-case(failure): `kubemarine install --without-act`
files: [dry-run-install-fail.zip](https://github.com/Netcracker/KubeMarine/files/11863981/dry-run-install-fail.zip)

test-case(success): `kubemarine install --without-act`
files: [dry-run-install.zip](https://github.com/Netcracker/KubeMarine/files/11864019/dry-run-install.zip)

test case(success): `kubemarine upgrade procedure.yaml --without-act`
files: [dry-run-upgrade.zip](https://github.com/Netcracker/KubeMarine/files/11863949/dry-run-upgrade.zip)

ilia1243:
The approach is ok, working on the implementation details.

kubemarine/core/flow.py
```diff
@@ -184,9 +184,8 @@ def run_tasks(resources: res.DynamicResources, tasks, cumulative_points=None, ta
     cluster = resources.cluster()
 
     if args.get('without_act', False):
+        cluster.context["dry_run"] = True
```
ilia1243, line 187: I suggest to not recreate the inventory on the deployer. For example, if you run upgrade, the input cluster.yaml will be changed. This does not seem as "dry-run". I suggest to create separate file in dump/ and reflect the changes there.

kubemarine/core/group.py
```diff
@@ -399,13 +428,16 @@ def _make_result_or_fail(self, results: _HostToResult,
         return group_result
 
     @_handle_internal_logging
+    @_handle_dry_run
     def run(self, *args, **kwargs) -> Union[NodeGroupResult, int]:
```
ilia1243, line 432: I suggest to not call NodeGroup.run/sudo/put for RW operations in dry run at all. The reason is it is difficult to simulate its behaviour. For example, you return empty results, but in fact it returns not empty results in real run. So this can lead to unpredictable changes in behaviour of Kubemarine. Also, put does not return result at all, run/sudo might return int in some cases, and so on...

ilia1243:
All procedure should be supported. If it is impossible, please introduce new option --dry-run and specify which procedures support it and which do not. Currently, --without-act is supported by all procedures except `migrate_kubemarine` and `do`.

ilia1243:
If it is possible, please make unit tests that will run the procedures in dry-run.

kubemarine/plugins/__init__.py
```diff
@@ -1018,14 +1029,14 @@ def apply_source(cluster: KubernetesCluster, config: dict) -> None:
...
         apply_common_group = cluster.create_group_from_groups_nodes_names(apply_groups, apply_nodes)
 
-    destination_common_group.put(source, destination_path, backup=True, sudo=use_sudo)
+    destination_common_group.put(source, destination_path, backup=True, sudo=use_sudo, dry_run=dry_run)
 
     if apply_required:
         method = apply_common_group.run
         if use_sudo:
             method = apply_common_group.sudo
         cluster.log.debug("Applying yaml...")
```
ilia1243, line 1038: Minor moment, rendering of plugins' templates might depend on previous shell commands. See `runtime_vars`. If it is possible, I suggest to skip rendering of templates in such cases, because its result will likely be not the same as during the real run.

koryaga:
Need to complete rework

PR #468, opened 2023-07-04
[MANOPD-90366] Improvements in remote executor and logging

ilia1243:
### Description
* Miscellaneous improvements in RemoteExecutor / NodeGroup API and internal behavior. Rework logging of errors / commands output.

### Solution
* Implemented callbacks in RemoteExecutor. Added CollectorCallaback to collect runners results from few commands in deferred mode.
* Support merging of few commands in RemoteExecutor with `warn=True`.
* Support workarounds in RemoteExecutor. Workarounds logic is moved from NodeGroup to RemoteExecutor.
* Remove `logging_stream` property from NodeGroup. Instead, `hide=False` now streams to logging framework.
    * No longer need to redirect stderr to stdout at process level.
* Timeouts for NodeGroup and DeferredGroup now behave the same:
    * If timeout is explicitly specified in run/sudo, it is applied only to Connection.
    * RemoteExecutor has global timeout for concurrent.future for each batch of commands.
    * Commands with explicit timeout can no longer be merged.
* Added custom UnexpectedExit and CommandTimedOut exceptions instead of that from `invoke`.
* Do not print command separator when logging of UnexpectedExit exception from RemoteExecutor.
* Rework logging of output. Now runners results are printed in the same format in all cases.
* Major rework in printing of errors.
* Generic results from commands execution can no longer be merged. Only runners results can be merged now.

### Test Cases

**TestCase 1**

1. Configure inventory with wrong configurition. For example, use `containerd` and specify
   ```yaml
   services:
     packages:
       associations:
         containerd:
           service_name: "not-a-containerd"
   ```
2. Run `install` procedure.

Results:

| Before | After |
| ------ | ------ |
| Error message contains symbols like `_-=____-_-__-___=-_---=-=______-` | Error message is more pretty |

### Checklist
- [x] Integration CI passed

#### Unit tests
test_group.py / test_executor.py / test_workaround.py - fixes to adapt to new API, and new tests for new features.

### DIFF
--- a/kubemarine/procedures/upgrade.py
+++ b/kubemarine/procedures/upgrade.py
@@ -152,7 +152,6 @@ def upgrade_containerd(cluster: KubernetesCluster):
                     os_specific_associations = cluster.get_associations_for_node(node.get_host(), 'containerd')
                     node.put(StringIO(config_string), os_specific_associations['config_location'],
                              backup=True, sudo=True, mkdir=True)
-                    print("Running !!!!!!!!!!!!!")
                     node.sudo(
                         f"sudo systemctl restart {os_specific_associations['service_name']} && "
                         f"systemctl status {os_specific_associations['service_name']} && " 
--- a/kubemarine/procedures/upgrade.py
+++ b/kubemarine/procedures/upgrade.py
@@ -152,9 +152,10 @@ def upgrade_containerd(cluster: KubernetesCluster):
                     os_specific_associations = cluster.get_associations_for_node(node.get_host(), 'containerd')
                     node.put(StringIO(config_string), os_specific_associations['config_location'],
                              backup=True, sudo=True, mkdir=True)
+                    print("Running !!!!!!!!!!!!!")
                     node.sudo(
                         f"sudo systemctl restart {os_specific_associations['service_name']} && "
-                        f"systemctl status {os_specific_associations['service_name']} &&" 
+                        f"systemctl status {os_specific_associations['service_name']} && " 
                         f"sudo systemctl restart kubelet",
                         callback=collector)
             return collector.result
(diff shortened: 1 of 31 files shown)


test/unit/core/test_group.py
```diff
@@ -27,14 +26,24 @@ def setUp(self):
...
+            \tsudo: kubectl: command not found
+            \t
+            10.101.1.3: code=1
+            \t=== stderr ===
+            \tsudo: kubectl: command not found
+            \t
+            10.101.1.4: code=1
+            \t=== stderr ===
+            \tsudo: kubectl: command not found
+            \t"""
```
ilia1243, line 41: I changed logging format of commands results. Previously it was
```
	10.102.0.133: code=0
		STDOUT: line 1
		        line 2
```
Now it is
```
10.101.17.41: code=0
	=== stdout ===
	line 1
	line 2
```
This is to make it similar to the format that is printed if UnexpectedExit occurs where `str(invoke.Result)` is typically printed. See https://github.com/pyinvoke/invoke/blob/1.6.0/invoke/runners.py#L1473

Please be free to comment on this and ask me to return the previous format if you find it more pretty.

kubemarine/procedures/upgrade.py
```diff
@@ -152,11 +152,12 @@ def upgrade_containerd(cluster: KubernetesCluster):
                     os_specific_associations = cluster.get_associations_for_node(node.get_host(), 'containerd')
                     node.put(StringIO(config_string), os_specific_associations['config_location'],
                              backup=True, sudo=True, mkdir=True)
-                    tokens.append(node.sudo(
+                    node.sudo(
                         f"sudo systemctl restart {os_specific_associations['service_name']} && "
-                        f"systemctl status {os_specific_associations['service_name']}"))
-                    tokens.append(node.sudo("systemctl restart kubelet"))
-            return exe.get_merged_runners_result(tokens)
```
koryaga, line 159: Reworked a bit during rebase

koryaga (approved):

heretic449 (approved):

PR #472, opened 2023-07-14
Add chcon command in function install(group) and setup_modprobe(group)

krang404:
### Description
When you install Kubernetes on a system with SELinux 3.5 version, you have a problem... kubelet.service doesn`t start. It happens because the file of unit kubelet hasn`t context **systemd_unit_file_t**  in attributes. Also after a reboot of any nodes of the cluster module **br_netfilter** is absent for the same reason - file in /etc/modules-load.d/ hasn`t required context.

### Solution
In function **install(group)** in kubemarine/kubemarine/\_\_init\_\_.py, in line 338, add command to change context from **user_tmp_t**  to **systemd_unit_file_t**. Likewise in function setup_modprobe(group) in kubemarine/system.py, in line 503, add change context from temp to *etc_t*. This solution has backward compatible with the previous version because this flag used to all of the system unit files in SELinux.  

### Tests in OS with different versions of SELinux
 - OS Centos 7.9 package policycoreutils-2.5
 - OS Oracle Linux 9.1 package policycoreutils-3.4
 - OS Oracle Linux 9.2 package policycoreutils-3.5

### Checklist
- [ ] Integration CI passed

### DIFF
--- a/kubemarine/kubernetes/__init__.py
+++ b/kubemarine/kubernetes/__init__.py
@@ -146,7 +146,7 @@ def enrich_inventory(inventory: dict, _):
             if "labels" not in node:
                 node["labels"] = {}
             node["labels"]["node-role.kubernetes.io/worker"] = "worker"
-            
+
     # Validate the provided podSubnet IP address
     pod_subnet = inventory.get('services', {}).get('kubeadm', {}).get('networking', {}).get('podSubnet')
     try:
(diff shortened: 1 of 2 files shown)


koryaga:
Which exact OS version do you use ?

krang404:
> Which exact OS and k8s version do you use ?

Oracle Linux 9.2

krang404:
This hotfix isn`t correct.

PR #474, opened 2023-07-19
[MANOPD-90205] fix for vulnerability scanning failure

pranavcracker:
### Description
* When we run the kube-bench on cluster running with k8s v1.27.1 installed using kubemarine 0.18.2, we got below failures-

```
[FAIL] 1.1.12 Ensure that the etcd data directory ownership is set to etcd:etcd (Automated)
```

Fixes # (issue)
**MANOPD-90205**

### Solution
* Fix for 1.1.12 - Set to ownership for data directory(/var/lib/etcd) as etcd:etcd 

### Test Cases

**TestCase 1**

* Find the latest version of kube-becnh utility [https://github.com/aquasecurity/kube-bench/releases](url) 
* Upload archive to the master/worker node, for example
`scp -i ~/.ssh/shift_test_key ~/kube-bench_0.6.15_linux_amd64.tar.gz ubuntu@:/home/ubuntu`
 
* Unpack it to the separate folder
`tar -xvf kube-bench_0.6.15_linux_amd64.tar.gz`

* Run check as (latest Kubernetes Benchmark will be used by default [https://github.com/aquasecurity/kube-bench/blob/main/docs/platforms.md#cis-kubernetes-benchmark-support])
`./kube-bench --config-dir <pwd>/cfg --config <pwd>/cfg/config.yaml`

Results:

| Before | After |
| ------ | ------ |
| [FAIL] 1.1.12 Ensure that the etcd data directory ownership is set to etcd:etcd (Automated) | [PASS] 1.1.12 Ensure that the etcd data directory ownership is set to etcd:etcd (Automated) |

### Checklist
- [x] Integration CI passed

### DIFF
--- a/kubemarine/procedures/install.py
+++ b/kubemarine/procedures/install.py
@@ -139,6 +139,19 @@ def system_prepare_system_sysctl(group: NodeGroup):
         sysctl.reload,
     ])
 
+@_applicable_for_new_nodes_with_roles('all')
+def system_prepare_add_etcd_user(group: NodeGroup):
+    cluster: KubernetesCluster = group.cluster
+    log = cluster.log
+    log.debug("Checking existence of etcd user for ETCD")
+    kubernetes_nodes = cluster.make_group_from_roles(['control-plane'])
+    for member_node in kubernetes_nodes.get_ordered_members_list():
+        user_exists = member_node.sudo("id etcd", warn=True)
+        # log.debug(user_exists.stderr_contains("no such user"))
+        if user_exists.stderr_contains("no such user"):
+            log.debug("etcd user doesn't exists, creating the user")
+            member_node.sudo("adduser etcd --disabled-password --gecos ''")
+        else: log.debug("etcd user already exits")
 
 @_applicable_for_new_nodes_with_roles('all')
 def system_prepare_system_setup_selinux(group: NodeGroup):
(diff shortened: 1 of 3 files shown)


kubemarine/procedures/install.py
```diff
@@ -139,6 +139,19 @@ def system_prepare_system_sysctl(group: NodeGroup):
         sysctl.reload,
     ])
 
+@_applicable_for_new_nodes_with_roles('all')
+def system_prepare_add_etcd_user(group: NodeGroup):
```
alexarefev, line 143: I think, these changes make sense if run ETCD container as non root user.

ilia1243: Checked installation with the changes from this PR, and etcd container is still running as root. Do I correctly understand that it is useless to change ownership of the directory in this case? @pranavcracker Could you shed more lights on what is the reason for changing of the ownership and under what conditions? Also found similar issue with comment https://github.com/rancher/rancher/issues/29311#issuecomment-785699317

pranavcracker: We haven't made any changes that will affect the container user, thus that is true. All that has changed is a new user named etcd has been created and the ownership of the data directory, which has been changed to `etcd:etcd`. In my opinion, there is no other result if we alter the ownership of these directories other than making the scan successful.

ilia1243: In this case I suggest to leave this as known issue and do not fix.

alexarefev: We could go back to the solution after the following enhancement implementation [run control-plane as non-root](https://github.com/kubernetes/kubeadm/issues/2473)

pranavcracker:
As with current solution, we are running the etcd containers with root user and hence we will consider this as known issue with our solution.
Closing this PR as no further action is required.

PR #475, opened 2023-07-20
patch for kubeadm_flags

pranavcracker:
### Description
* Patch for updating --pod-infra-container-image version in /var/lib/kubelet/kubeadm-flags.env file

Fixes # (issue)
* MANOPD-88293

### Solution
* Apply this patch for the old clusters where --pod-infra-container-image version was not updated in /var/lib/kubelet/kubeadm-flags.env file even after k8s upgrade.

### How to apply
* [Install task/s](https://github.com/Netcracker/KubeMarine/blob/main/documentation/Maintenance.md#kubemarine-migration-procedure)

### Test Cases

**TestCase 1**
Steps:
1. Try to upgrade k8s for e.g. from _v1.24.11_ to _v1.25.2_ or _v1.25.2_ to _v1.26.4_ (where pause image version should be new between the versions) using tag 0.18.2 or older.
2. With such upgrade, the pause version for ` --pod-infra-container-image` will not get updated in the `/var/lib/kubelet/kubeadm_falgs.env` file.
4. Then apply this patch runing` kubemarine migrate_kubemarine` command.
5. Check the `/var/lib/kubelet/kubeadm_falgs.env` file to verify if   pause version is updated for ` --pod-infra-container-image` as per the k8s version.

Note - For checking the mapping between k8s version and pause image version please refer to the below document
https://github.com/Netcracker/KubeMarine/blob/0.18.2/kubemarine/resources/configurations/compatibility/internal/kubernetes_images.yaml#L135

### DIFF
--- a/kubemarine/patches/p1_kubeadm_flags.py
+++ b/kubemarine/patches/p1_kubeadm_flags.py
@@ -1,0 +1,44 @@
+from textwrap import dedent
+from io import StringIO
+
+from kubemarine.core.action import Action
+from kubemarine.core.patch import RegularPatch
+from kubemarine.core.resources import DynamicResources
+from kubemarine import kubernetes
+
+class TheAction(Action):
+    def __init__(self):
+        super().__init__("Updpate kubeadm_flags after cri upgrade")
+
+    def run(self, res: DynamicResources):
+        cluster = res.cluster()
+        path = 'plugins."io.containerd.grpc.v1.cri"'
+        target_kubernetes_version = cluster.inventory["services"]["kubeadm"]["kubernetesVersion"]
+        kubernetes_nodes = cluster.make_group_from_roles(['control-plane', 'worker'])
+        pause_version = cluster.globals['compatibility_map']['software']['pause'][target_kubernetes_version]['version']
+        sandbox = cluster.inventory["services"]["cri"]['containerdConfig'][path]["sandbox_image"]
+        param_begin_pos = sandbox.rfind(":")
+        sandbox = sandbox[:param_begin_pos] + ":" + str(pause_version)
+        for member_node in kubernetes_nodes.get_ordered_members_list():
+            kubeadm_flags_file = "/var/lib/kubelet/kubeadm-flags.env"
+            kubeadm_flags = member_node.sudo(f"cat {kubeadm_flags_file}").get_simple_out()
+            updated_kubeadm_flags = kubernetes._config_changer(kubeadm_flags, f"--pod-infra-container-image={sandbox}")
+            member_node.put(StringIO(updated_kubeadm_flags), kubeadm_flags_file, backup=True, sudo=True)
+
+
+
+class KubeadmFlags(RegularPatch):
+    def __init__(self):
+        super().__init__("kubeadm_flags")
+
+    @property
+    def action(self) -> Action:
+        return TheAction()
+
+    @property
+    def description(self) -> str:
+        return dedent(
+            f"""\
+            Patch to update --pod-infra-contianer-image version after upgrading the k8s version.
+            """.rstrip()
+        )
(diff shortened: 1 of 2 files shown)


kubemarine/patches/p1_kubeadm_flags.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self):
+        super().__init__("Updpate kubeadm_flags after cri upgrade")
+
+    def run(self, res: DynamicResources):
+        cluster = res.cluster()
+        path = 'plugins."io.containerd.grpc.v1.cri"'
+        target_kubernetes_version = cluster.inventory["services"]["kubeadm"]["kubernetesVersion"]
+        kubernetes_nodes = cluster.make_group_from_roles(['control-plane', 'worker'])
+        pause_version = cluster.globals['compatibility_map']['software']['pause'][target_kubernetes_version]['version']
+        sandbox = cluster.inventory["services"]["cri"]['containerdConfig'][path]["sandbox_image"]
```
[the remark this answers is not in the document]
pranavcracker: added a condition to check if `sandbox_image` is specified in `cluster.yaml`. If it is present, patch will be skipped with error message.

kubemarine/patches/p1_kubeadm_flags.py
```diff
@@ -0,0 +1,62 @@
...
+        param_begin_pos = sandbox.rfind(":")
+        sandbox = sandbox[:param_begin_pos] + ":" + str(pause_version)
+        if "cri" in cluster.raw_inventory["services"] and "sandbox_image" in cluster.raw_inventory["services"]["cri"]["containerdConfig"][path]:
+            print("Skipping the patch for Kubeadm_Flags")
+            exit(0)
```
ilia1243, line 38: 1. Custom inventory may contain `services.cri` section, but not `services.cri.containerdConfig.plugins."io.containerd.grpc.v1.cri"`. In this case it will fail.
2. Please do not `exit`. See https://github.com/Netcracker/KubeMarine/blob/0.19.0/kubemarine/core/action.py#L40
3. Please use logging approach instead of `print`.

kubemarine/patches/p1_kubeadm_flags.py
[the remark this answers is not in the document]
ilia1243: Thank you, but p. 1 is still actual.

kubemarine/patches/p1_kubeadm_flags.py
pranavcracker: 1. Added the condition to handle if inventory do not contain `services.cri.containerdConfig.plugins."io.containerd.grpc.v1.cri"`.
2. Also updated error msg to use `cluster.log.debug` instead of `print`

ilia1243 (approved):

kubemarine/patches/p1_kubeadm_flags.py
```diff
@@ -0,0 +1,70 @@
...
+
+            cri_config = cluster.raw_inventory.get("services", {}).get("cri", {})
+            containerd_config = cri_config.get("containerdConfig", {})
+            sandbox_image = containerd_config.get(path, {}).get("sandbox_image")
+            if sandbox_image is None:
+                for member_node in kubernetes_nodes.get_ordered_members_list():
+                    kubeadm_flags_file = "/var/lib/kubelet/kubeadm-flags.env"
+                    kubeadm_flags = member_node.sudo(f"cat {kubeadm_flags_file}").get_simple_out()
+                    updated_kubeadm_flags = kubernetes._config_changer(kubeadm_flags, "--pod-infra-container-image=%s" %sandbox )
+                    member_node.put(StringIO(updated_kubeadm_flags), kubeadm_flags_file, backup=True, sudo=True)
```
ilia1243, line 48: The same problem as in original PR: files is updated in filesystem, but the changes are not picked up and kubelet might be still running with old value of the flag.

pranavcracker: added command to restart kubelet service after updating kubeadm flags.

ilia1243 (approved):

ilia1243:
Please also write Test Case for Q/A how to emulate the situation when the patch changes the state of the cluster

PR #487, opened 2023-08-03
PSUPCLPL-13194 Optimize downloading of Kubernetes resources during backup

ilia1243:
### Description
* `export.kubernetes` task of `backup` procedure works long.

### Solution
* Use one `kubectl` invocation to get all resources for the particular namespace, instead of multiple `kubectl && kubectl` invocations.
    * Implement custom parsing of the resulting yaml list to split resources of different kinds into different files.
* Use 2 background tasks for all available control planes to invoke few `kubectl` commands in parallel.
* Use `gzip` to compress data before sending through the network.
* Use dedicated parsing background task to ensure that only one thread loads the data in memory.

### Test Cases

**TestCase 1**

1. Run `backup` or `backup --tasks export.kubernetes,pack` on large cluster

ER: elapsed time for the `export.kubernetes` task decreases 10 times or more.

### Checklist
- [x] Integration CI passed

#### Unit tests
test_backup.py - added test for `export.kubernetes` task.

### DIFF
--- a/kubemarine/core/utils.py
+++ b/kubemarine/core/utils.py
@@ -15,6 +15,7 @@ import hashlib
 import io
 import json
 import os
+import re
 import shutil
 import sys
 import time
@@ -495,10 +496,24 @@ def parse_aligned_table(table_text: str) -> List[Dict[str, str]]:
     :return: List of rows
     """
     rows = table_text.strip().split('\n')
-    headers = rows[0].split()
-    headers_num = len(headers)
-    headers_pos = [rows[0].index(h) for h in headers]
 
+    # Parse headers and their positions. No leading or trailing spaces are expected in the line.
+    headers_line = rows[0]
+    headers = []
+    headers_pos = []
+    pos = 0
+    for match in re.finditer(r'\s+', headers_line):
+        headers_pos.append(pos)
+        span = match.span()
+        headers.append(headers_line[pos:span[0]])
+        pos = span[1]
+
+    headers_pos.append(pos)
+    headers.append(headers_line[pos:len(headers_line)])
+
+    # Parse each row in the table,
+    # provided that the columns start at the same positions as the headers
+    headers_num = len(headers)
     data = []
     for row_text in rows[1:]:
         row = {}
--- a/kubemarine/core/utils.py
+++ b/kubemarine/core/utils.py
@@ -20,7 +20,7 @@ import sys
 import time
 import tarfile
 
-from typing import Tuple, Callable, List, TextIO, cast, Union, TypeVar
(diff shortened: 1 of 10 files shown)


kubemarine/procedures/backup.py
```diff
@@ -252,136 +257,387 @@ def export_kubernetes_version(cluster: KubernetesCluster) -> None:
...
+            val = yaml.safe_load(item_prop)[key]
+
+        return val
+
+    def _handle(self, payload: ParserPayload) -> None:
+        namespace = payload.namespace
+        if namespace:
+            self.logger.debug(f"Loading resources from namespace {namespace!r}...")
+        else:
+            self.logger.debug(f"Loading non-namespaced resources...")
```
Imadzuma, line 402: After running `export.kubernetes` task there are full mesh in logs on verbose level (because every thread writes it's own logs at the same time). Maybe we will add thread id to log messages to handle logs order on different threads?
Additionally, if some kubectl command was unsuccessful on some namespace, we can't understand, on which namespace it was. I think, additional logs are needed in case of exception.

kubemarine/core/utils.py
```diff
@@ -486,6 +486,35 @@ def _test_version(version: str, numbers_amount: int) -> List[int]:
...
+    Parse aligned table into the list of {header: cell} map.
+    The text with table can be initially produced, for example, by https://pkg.go.dev/text/tabwriter.
+
+    :param table_text: aligned table as string
+    :return: List of rows
+    """
+    rows = table_text.strip().split('\n')
+    headers = rows[0].split()
+    headers_num = len(headers)
+    headers_pos = [rows[0].index(h) for h in headers]
```
Imadzuma, line 500: Do you guarantee, that no header can be substring for another one?

kubemarine/procedures/backup.py
ilia1243: > Maybe we will add thread id to log messages to handle logs order on different threads?

Added

> if some kubectl command was unsuccessful on some namespace, we can't understand, on which namespace it was

There will be an error: `Unexpected exit, command: '... -n <namespace >...'`, but I added one more explicit logging.

kubemarine/core/utils.py
ilia1243: Thank you for pointing out this gap. Although this code much relies on format, for example it assumes that there is no spaces in header names, this particular gap can be fixed. Now both header name and positions are detected in one pass.

Imadzuma (approved):

heretic449 (approved):

koryaga (approved):

PR #519, opened 2023-10-12
[PSUPCLPL-13705] Certificates_not_updated_on_workers_nodes

nikhil1697:
### Description
Modify the k8s_certs_renew_task function to perform certificate renewal on both control plane nodes and worker nodes when required.

Fixes # (issue)
The primary fix in this change is to ensure that both control plane and worker nodes have their certificates kubelet.crt renewed.

### Solution
The function k8s_certs.force_renew_kubelet_serving_certs is called on the worker nodes. This function focuses on renewing the kubelet serving certificates on worker nodes.

### Test Cases

**TestCase 1**

Steps:

1. Run kubemarine cert_renew on a k8s cluster

Results:

| Before | After |
| ------ | ------ |
| kubelet.crt & kubelet.key was updated only on master nodes | kubelet.crt & kubelet.key updated on all nodes |

### Checklist
- [ ] Integration CI passed

### DIFF
--- a/kubemarine/procedures/cert_renew.py
+++ b/kubemarine/procedures/cert_renew.py
@@ -57,6 +57,7 @@ def k8s_certs_renew_task(cluster: KubernetesCluster) -> None:
 
     cluster.log.debug("Starting certificate renewal for kubernetes")
     cluster.nodes['control-plane'].call(k8s_certs.renew_apply)
+    cluster.nodes['worker'].call(k8s_certs.force_renew_kubelet_serving_certs)
 
 
 def k8s_certs_overview_task(cluster: KubernetesCluster) -> None:


kubemarine/procedures/cert_renew.py
```diff
@@ -57,6 +57,7 @@ def k8s_certs_renew_task(cluster: KubernetesCluster) -> None:
 
     cluster.log.debug("Starting certificate renewal for kubernetes")
     cluster.nodes['control-plane'].call(k8s_certs.renew_apply)
+    cluster.nodes['worker'].call(k8s_certs.force_renew_kubelet_serving_certs)
```
alexarefev, line 60: I think the code changes should be implemented in `k8s_certs.py`

ilia1243:
Hello. I see that we already renew the cert for control planes, but do not understand the reason for that. As far as I understood from some internet discussions, this serving cert is self-signed and no one trust it by default. It means that if someone connects to kubelet through port 10250, it has to skip cert verification. This in particular skips verification of expiration date. This implies that there is no use to renew it. On the other hand, to make connection pass the verification, it is suggested to configure https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs this will make custom renew procedure useless and maybe even harmful.

koryaga:
To cover this issue along with CPDEV-93819

PR #525, opened 2023-10-17
[CPDEV-93091] Add dashboard ingress check to check_paas dashboard test

n549:
### Description
During `check_paas` tests it would be useful to check not only `kubernetes-dashboard` service, but also `kubernetes-dashboard` ingress.

Fixes # (issue)
CPDEV-93091

### Solution
`kubernetes_dashboard_status` function is updated with the check of the ingress similar way as it was done for the service.
The ingress is being checked only after successful check of the service.

To make the ingress check possible in an environment without proper DNS settings, the curl's `--resolve` flag is used.
The ingress name is being resolved to the cluster.inventory['control_plain']['internal'] and the check is running from one of master nodes.

### How to apply
Not applicable

### Test Cases
1. Deploy a cluster with `kubernetes-dashboard` plugin. Run `check_paas` procedure with `--tasks=kubernetes.plugins.dashboard` flag.

ER: The test is successful, debug output show 2 sub-checks: one of the service and one of the ingress.

### Checklist
- [x] Integration CI passed

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -75,17 +75,6 @@ jobs:
       - name: Check paas
         id: test_check_paas
         run: kubemarine check_paas -c ${{ env.cluster_yaml_file }} --dump-location ./results/check_paas_dump/
-      - name: Check ingress connectivity
-        run: |
-          sudo echo "172.17.0.1 dashboard.test-local-k8s.com" | sudo tee -a /etc/hosts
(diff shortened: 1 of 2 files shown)


ilia1243:
I think checks `name: Check ingress connectivity` can be removed now from https://github.com/Netcracker/KubeMarine/blob/main/.github/workflows/integration_tests.yaml

n549:
> I think checks `name: Check ingress connectivity` can be removed from https://github.com/Netcracker/KubeMarine/blob/main/.github/workflows/integration_tests.yaml as they are covered by `check_paas` now.

Done.

koryaga (approved):

PR #523, opened 2023-10-16
[CPDEV-93827] - add new containerd format support

Imadzuma:
### Description
Now we support registry configuration for containerd using `registry.configs` and `registry.mirrors` in `/etc/containerd/config.toml` (parameter `containerdConfig` from cluster.yaml):
https://github.com/Netcracker/KubeMarine/blob/0.22.0/documentation/Installation.md#cri

But this approach is deprecated from containerd v1.5.0: https://github.com/containerd/containerd/blob/v1.5.0/docs/cri/registry.md#configure-image-registry
In kubemarine we support versions v1.5.* and v1.6.*, for this reason we can fully migrate to new approach for now.

Additionally, the current approach is not compatible with ctr tool unlike the new one.

More about the new approach can be found in doc: https://github.com/containerd/containerd/blob/v1.5.1/docs/hosts.md

### Solution
Although `registry.configs` is deprecated, there is no alternative for credentials configuartion: it can be specified using `registry.configs.auth`. For this reason, we are going to support new configuration format and recommend to use it in our documents. But we don't restrict to use old format and don't provide migration patch:
* added new `containerdRegistriesConfig` field, that is used to specify registries hosts configuration format;
* added `plugins."io.containerd.grpc.v1.cri".registry.config_path` enrichment, if no old format configuration is specified;
* added exception, if specified field in both formats;
* registries enrichment is migrated to new format, if it's possible;
* added regular patch to move registries to new format;
* `containerdRegistriesConfig` is included in check_paas procedure;

Little fixes:
* Some symbols are restricted in windows/macOS OSes, but we can use those symbols in dump files. It follows not expected behavior. Those symbols were replaced with '_' symbol. See test cases for examples.
* Sometimes restore procedure fails because of failing etcd container removing using ctr: root cause in async `rm task` command, that can not be finished, when this task should be removed: added removing with force;

### How to apply
Provide steps how to apply on top previous Kubemarine version to execute on existing clusters
* [Install task/s](documentation/Installation.md#installation-tasks-description)
* Manual step 
* [Upgrade procedure](documentation/Maintenance.md#upgrade-procedure)

### Test Cases
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

**TestCase 1**: exception, if `containerdRegistriesConfig` is combined with old format fields.

Test Configuration:

- Hardware: 
- OS: 
[...]

PR #531, opened 2023-10-24
[CPDEV-94111] - fix ctr flags for new containerd format

Imadzuma:
### Description
This PR fixes the bug, related to PR https://github.com/Netcracker/KubeMarine/pull/523:
If user specifies only new format configuration (without credentials), `hosts-dir` flag is not applied for ctr commands.
Example of configuration:
```yaml
services:
  kubeadm:
    imageRepository: registry.example:9090
  cri:
    containerdRegistriesConfig:
      registry.example:9090:
        host."http://registry.example:9090":
          capabilities: [ "pull", "resolve" ]
          skip_verify: true
```
As result ctr fails with pulling problem like:
```
failed to do request: Head \"https://registry.example:9090/v2/pause/manifests/3.9\": http: server gave HTTP response to HTTPS client" host="registry.example:9090"
ctr: failed to resolve reference "registry.example:9090/pause:3.9": failed to do request: Head "https://registry.example:9090/v2/pause/manifests/3.9": http: server gave HTTP response to HTTPS client
```
The root cause in ctr_flags.conf that saves ctr options per registry in old format, but --hosts-dir should be applied for any registry, if the new format is used. 

Fixes # (issue)

### Solution
* Added `*` rule in ctr_flags.conf, that means, that options should be applied for any registry;
* Fixed `etcdctl`, that can parse '*' rules in ctr_flags.conf;
* If `config-path` is enabled, `*=--hosts-dir <config_path>` is added to  ctr_flags.conf;
* Added patch to reinstall etcdctl;
* Fixed reconfigure_registry patch using deviated_cluster, because it's more safe way to compare configuration differences when `raw_inventory`;

### Test Cases

**TestCase 1**

Test Configuration:

- Hardware: 
- OS: 
- Inventory: containerd configuration like from description above

Steps:

1. Kubemarine install
2. Kubemarine check_paas

Results:

| Before | After |
| ------ | ------ |
| `etcd.health_status` failed with error: ctr: image "registry.example:9090/etcd:3.5.6-0": not found | `etcd.health_status` task works correctly |

### Checklist
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/kubemarine/core/defaults.py
+++ b/kubemarine/core/defaults.py
@@ -171,7 +171,8 @@ def apply_registry(inventory: dict, cluster: KubernetesCluster) -> dict:
             containerd_endpoints = ["%s://%s" % (protocol, registry_mirror_address)]
 
         old_format_result, _ = contains_old_format_properties(inventory)
-        if old_format_result:
+        # todo remove p3_reconfigure_registries after next release
+        if old_format_result or (cluster and not cluster.context.get('p3_reconfigure_registries', True)):
             # Add registry info in old format
             registry_section = f'plugins."io.containerd.grpc.v1.cri".registry.mirrors."{registry_mirror_address}"'
             containerd_config = inventory['services']['cri']['containerdConfig']
--- a/kubemarine/core/defaults.py
+++ b/kubemarine/core/defaults.py
@@ -172,7 +172,7 @@ def apply_registry(inventory: dict, cluster: KubernetesCluster) -> dict:
 
         old_format_result, _ = contains_old_format_properties(inventory)
         # todo remove p3_reconfigure_registries after next release
-        if old_format_result or (cluster and not cluster.context.get('p3_reconfigure_registries', True)):
+        if old_format_result or not cluster.context.get('p3_reconfigure_registries', True):
             # Add registry info in old format
             registry_section = f'plugins."io.containerd.grpc.v1.cri".registry.mirrors."{registry_mirror_address}"'
             containerd_config = inventory['services']['cri']['containerdConfig']
--- a/kubemarine/cri/containerd.py
+++ b/kubemarine/cri/containerd.py
@@ -277,7 +277,7 @@ def configure_crictl(group: NodeGroup) -> None:
 
 
 def get_config_path(inventory: dict) -> Optional[str]:
-    config_path: Optional[str] = inventory.get('containerdConfig', {}) \
+    config_path: Optional[str] = inventory.get('services', {}).get('cri', {}).get('containerdConfig', {}) \
         .get('plugins."io.containerd.grpc.v1.cri".registry', {}).get('config_path')
     return config_path
 
--- a/kubemarine/cri/containerd.py
+++ b/kubemarine/cri/containerd.py
@@ -256,10 +256,11 @@ def configure_ctr_flags(group: NodeGroup) -> None:
         elif registry_auth.get('username'):
             options.append(f'--user {registry_auth["username"]}' +
                            f':{registry_auth["password"]}' if registry_auth.get("password") else '')
-        # Add hosts-dir, if it's presented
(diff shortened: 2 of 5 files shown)


kubemarine/cri/containerd.py
```diff
@@ -256,11 +256,12 @@ def configure_ctr_flags(group: NodeGroup) -> None:
         elif registry_auth.get('username'):
             options.append(f'--user {registry_auth["username"]}' +
                            f':{registry_auth["password"]}' if registry_auth.get("password") else '')
-        # Add hosts-dir, if it's presented
-        if config_path:
-            options.append(f'--hosts-dir {config_path}')
```
ilia1243, line 261: Won't it be simpler if just iterate over `set().union(registry.get('mirrors', {}).keys(), registry.get('configs', {}).keys(), cluster.inventory.get("services", {}).get("cri", {}).get('containerdRegistriesConfig', {}))`, and return these lines? In this case we could avoid extra patch

kubemarine/core/defaults.py
```diff
@@ -171,7 +171,8 @@ def apply_registry(inventory: dict, cluster: KubernetesCluster) -> dict:
             containerd_endpoints = ["%s://%s" % (protocol, registry_mirror_address)]
 
         old_format_result, _ = contains_old_format_properties(inventory)
-        if old_format_result:
+        # todo remove p3_reconfigure_registries after next release
+        if old_format_result or (cluster and not cluster.context.get('p3_reconfigure_registries', True)):
```
ilia1243, line 175: bool(cluster) is always True. Also I suppose the same check should be applied to all cases where enrichment has changed. I remember that we also changed enrichment in containerd.enrich_inventory

kubemarine/patches/p3_reconfigure_registries.py
```diff
@@ -25,18 +25,18 @@ def __init__(self) -> None:
...
-        result, _ = contains_old_format_properties(raw_inventory)
-        if result:
-            logger.info("Found old format configuration, updating containerd configuration is not needed")
+
+        prev_cluster = res.create_deviated_cluster({
+            'p3_reconfigure_registries': False
+        })
+
+        if 'containerdRegistriesConfig' not in cluster.inventory['services']['cri'] or \
+                'containerdRegistriesConfig' in prev_cluster.inventory['services']['cri']:
```
ilia1243, line 38: Not sure that I understand these conditions. Is the second condition can really be True in some cases? Consider that 'containerdRegistriesConfig' cannot not be specified for old inventories. Maybe the first condition is enough and create_deviated_cluster is not necessary at all. Alternatively, the result of old and new enrichment can be compared by just `cluster.inventory['services']['cri'] == prev_cluster.inventory['services']['cri']`

kubemarine/cri/containerd.py
ilia1243: Discussed to maintain common flag for all registries through `*=...`

ilia1243 (approved):

kubemarine/core/defaults.py
Imadzuma: bool(cluster) was fixed.
Enrichment changes only for registries, if they are enriched in new format. In other cases nothing changes.

kubemarine/patches/p3_reconfigure_registries.py
Imadzuma: 'containerdRegistriesConfig' for `prev_cluster` checks, that patch is not applied after manual changes in cluster.yaml. In that case user should manually run containerd reconfiguration, but patch should do nothing.
In other cases `cluster.inventory['services']['cri'] == prev_cluster.inventory['services']['cri']` handles the same cases, for this reason it doesn't matter what is used in that condition.

ilia1243 (approved):

PR #547, opened 2023-11-14
[CPDEV-94948] CoreDNS resolving issues

alexarefev:
### Description
* Pod receives  `server can't find ...` response from CoreDNS
* Pod doesn't receives any response from CoreDNS the error is the following:
```
;; connection timed out; no servers could be reached
```

### Solution
* Update Troubleshooting guide.

### Checklist

#### Unit tests

### DIFF
--- a/documentation/Troubleshooting.md
+++ b/documentation/Troubleshooting.md
@@ -30,6 +30,9 @@ This section provides troubleshooting information for Kubemarine and Kubernetes 
   - [No Pod-to-Pod Traffic for Some Nodes with More Than One Network Interface](#no-pod-to-pod-traffic-for-some-nodes-with-more-than-one-network-interface)
   - [No Pod-to-Pod Traffic for Some Nodes with More Than One IPs with Different CIDR Notation](#no-pod-to-pod-traffic-for-some-nodes-with-more-than-one-ips-with-different-cidr-notation)
   - [Ingress Cannot Be Created or Updated](#ingress-cannot-be-created-or-updated)
+  - [CoreDNS Cannot Resolve the Name](coredns-cannot-resolve-the-name]
+    - [Case 1](case-1]
+    - [Case 2](case-2]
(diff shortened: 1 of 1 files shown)


documentation/Troubleshooting.md
```diff
@@ -927,6 +930,38 @@ spec:
...
+```
+$ nslookup kubernetes.default
+Server:         172.30.0.10
+Address:        172.30.0.10:53
+
+
+** server can't find kubernetes.default: NXDOMAIN
+```
+
+**Root cause**: Images that are based on Alpine Linux do not use `search` directives in `/etc/resolv.conf` by default.
```
koryaga, line 948: can we put an Alpine version which has such behaviour and find some explanation why: design intent or a bug/gap ?

alexarefev: Sure, I'll extend description

n549, line 948: It's not Alpine Linux image per se, it's busybox utility in the Alpine, who causes this error (for example, see https://github.com/docker-library/busybox/issues/48), so that may happen with other images too.

alexarefev: Could you give an example, please?
I'm going to add WA from: [Does Alpine have known DNS issue within Kubernetes?](https://stackoverflow.com/questions/65181012/does-alpine-have-known-dns-issue-within-kubernetes)
Maybe there is something else that I should mention

n549: Yes, this WA is fine for Alpine-based images.
Busybox images themselves also can have issues, for example, versions <=1.28.4 and >=1.35.0 resolve short names, but intermediate versions (for example, 1.29.3 or 1.33) give errors for existing names.

alexarefev: Thanks, I'll take it into account

n549 (approved):

koryaga (approved):

shmo1218 (approved):
The doc review is done.

PR #551, opened 2023-11-16
[CPDEV-94547] Rework network connectivity check and implement VRRP IP check

### LINKED ISSUE #513: Rework `network.check_tcp_ports` task
Now we check, if some ports are open on defined nodes: https://github.com/Netcracker/KubeMarine/blob/0.21.1/kubemarine/procedures/check_iaas.py#L867-L881
But we check the constant number of ports for all nodes including balancers. But required ports depend on node roles: https://kubernetes.io/docs/reference/networking/ports-and-protocols/

Additionally, after merging PR https://github.com/Netcracker/KubeMarine/pull/509 we will get additional required ports for worker nodes.

Need to rework this task to make checked ports be depended on node roles and support new ports here.

ilia1243:
### Description
* There are no VRRP IP connectivity checks.
* `network.check_tcp_ports` task of check_iaas procedure does not take into account roles of the nodes.
* There are some extra and missed ports to check connectivity. Also, UDP ports are not checked.

Fixes #513

### Solution
* Added `network.vips_connectivity` task that
   * assigns each VRRP IP to each balancer as alias to the host network
   * performs TCP ports connectivity check for ports that are necessary for balancers
* Added input/output mapping of ports for each node role. Connectivity is checked for all pairs of nodes (source, target) for which ports are matched in `output` and `input` sections correspondingly taking into account node roles of the source and target.
* `network.check_tcp_ports` is renamed to `network.ports_connectivity`.
* Added ports:
    * 20080/20443 depending on the schema.
    * 5443 for Calico API server
    * 53 (TCP & UDP) for CoreDNS
* Removed ports:
    * 10257, 10259, as they are bound to localhost and do not participate in inter-node connectivity.
    * 9094, 10254, as the connectivity is required only for pod network.
* Other notes:
    * TCP ports check is no longer skipped on installed cluster.
      Instead, simple pythonic client is implemented that only connects to services, but does not send the data.
    * Random NodePort in host network and random user port in pods subnet are checked instead of hardcoded ones.

### Test Cases

**TestCase 1**

Check VRRP IPs connectivity.

Steps:

1. Configure the infrastructure:
    * 2 balancers and 1 or more VRRP IP.
    * Different supported operating systems.
    * IPv4 or IPv6
    * Traffic is allowed or prohibited to balancers for ports 80, 443, 6443 through the VRRP IP.
2. Run `check_iaas --tasks network.vips_connectivity`

ER: The check succeeds if and only if the traffic is allowed. All balancers are checked.

3. Install the cluster (traffic should be allowed).
4. Disallow the traffic if necessary.
5. Run `check_iaas --tasks network.vips_connectivity`

ER: The check succeeds if and only if the traffic is allowed. Only one balancer is checked on which the VRRP IP is assigned.

**TestCase 2**

Strictest security rules.

Steps:

1. Configure the infrastructure with the strictest possible security.
    For example, prohibit 80 / 443 ports traffic on workers if there are balancers in the scheme.
    Take into account:
    * Different supported operating systems.
    * Different schemes (full-HA, minina-HA, scheme without balancers)
    * IPv4 or IPv6
2. Run `check_iaas --tasks network.ports_connectivity`

Results:

| Before | After |
| ------ | ------ |
| Check fails due to some ports are not opened | Check succeeds |

3. Install the cluster.
4. Run `check_iaas --tasks network.ports_connectivity`

| Before | After |
| ------ | ------ |
| Check is skipped | Check is performed and succeeded |

**TestCase 3**

UDP ports connectivity check.

Steps:

1. Disallow UDP traffic for:
    * All ports for pods subnet
    * 53 port for host network
2. Run `check_iaas --tasks network.pod_subnet_connectivity,network.ports_connectivity`

ER: 009 or 011 checks fail depending on what traffic is disallowed at step 1.

### Checklist
- [x] Integration CI passed


shmo1218 (approved):
The doc review is done.

alexarefev (approved):

PR #581, opened 2023-12-28
[CPDEV-96115]-Enabling_strong_crypto_support_by_default

nikhil1697:
### Description
These cipher suites are configuring the server to support a selection of strong and secure cryptographic algorithms for secure communication over TLS (Transport Layer Security).
* 

**Fixes # (issue)**
[CPDEV-96115](https://tms.netcracker.com/browse/CPDEV-96115)

**Security Strength:**
These cipher suites prioritize the use of strong cryptographic algorithms, enhancing the security strength of the TLS connections. They provide forward secrecy through the ECDHE key exchange.

**Cipher Suite Diversity:**
Including a mix of AES-GCM and ChaCha20-Poly1305 cipher suites ensures compatibility with a broader range of clients and provides alternatives in case certain algorithms are vulnerable or less secure.

### Solution

Enhanced Security:
The configured cipher suites enhance the security of the TLS communication by using modern cryptographic algorithms and key exchange methods.

Forward Secrecy:
Ephemeral key exchange (ECDHE) ensures forward secrecy, meaning that even if the server's private key is compromised, past communications remain secure.

Compatibility:
The chosen cipher suites strike a balance between security and compatibility. 

Performance:
ChaCha20-Poly1305 cipher suites are known for their good performance, which can be beneficial in resource-constrained environments.
* 

### Test Cases

**TestCase 1**

Test Configuration:

- OS: Ubuntu-20.04

Steps:

1. Installed a miniha cluster
2. Installed kube-bench to make KubeMarine CIS aligned. 
3. Run scan using command - kube-bench and check for identifier **1.2.31**.

Results:

| Before | After |
| ------ | ------ |
| Result used to be Warning | Result shows Pass |

**TestCase 2**
Apply the patch for the old clusters where API Server Cryptographic Ciphers are not update even after even after k8s upgrade.

Test Configuration:
Steps: 
1. Try to upgrade k8s for e.g. from v1.24.11 to v1.25.2 
2. With such upgrade --tls-cipher-suites on /etc/kubernetes/manifests/kube-apiserver.yaml will not get updated.
3. Then apply this patch runing kubemarine migrate_kubemarine command.
4. Check the /etc/kubernetes/manifests/kube-apiserver.yaml file to verify if --tls-cipher-suites is updated and also  run scan using command - kube-bench and check for identifier **1.2.31**.

Results:

| Before | After |
| ------ | ------ |
| Result used to be Warning | Result shows Pass |

**TestCase 3**
Apply the patch for the old clusters where API Server Cryptographic Ciphers are already available or some other ciphers are already present.

Results:
It will skip the patch by logging ciphers are already present.

### Checklist
- [ ] Integration CI passed


OleksandrIaresko:
The another one case is Kubernetes upgrade. 
Nothing was changed after upgrade scenario, the security report contains the same Warning
`[WARN] 1.2.30 Ensure that the API Server only makes use of Strong Cryptographic Ciphers (Manual)`

ilia1243:
Ideally, we have to write patch that makes the cluster and new resolved inventory consistent.
https://github.com/Netcracker/KubeMarine/blob/v0.25.1/documentation/internal/Patches.md

kubemarine/patches/strong_ciphers_patch.py
```diff
@@ -0,0 +1,52 @@
...
+
+            # Load the YAML configuration
+            try:
+                apiserver_config = yaml.safe_load(member_node.sudo(f"cat {apiserver_file}").get_simple_out())
+            except yaml.YAMLError as exc:
+                cluster.log.error(f"Failed to parse YAML file: {exc}")
+                return
+
+            # Modify the YAML structure (adjust the path as needed)
+            apiserver_config['spec']['containers'][0]['command'].append("--tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384")
```
ilia1243, line 30: Please check if this new arg is persisted after upgrade. If it does not, we have two known ways to reconfigure kube-apiserver manifest.
[install.deploy_kubernetes_audit](https://github.com/Netcracker/KubeMarine/blob/v0.25.1/kubemarine/procedures/install.py#L190) of install procedure and [admission.manage_pss](https://github.com/Netcracker/KubeMarine/blob/v0.25.1/kubemarine/admission.py#L645) of manage_pss procedure.

ilia1243, line 30: What if the manifest already has the `--tls-cipher-suite` arg? For example, someone installed the cluster using this guide https://github.com/Netcracker/KubeMarine/blob/main/documentation/internal/Hardening.md#automated-application-for-strong-cyrptographic-ciphers-for-api-server-during-new-cluster-installation

Also, probably the guide `Use strong cryptographic ciphers for API server` becomes no longer actual because we are starting to ensure corresponding parameters.

ilia1243, line 30: I suggest to take probable configuration `services.kubeadm.apiServer.extraArgs.tls-cipher-suites` into account. Probably someone required and already installed different set ciphers. I suggest to skip the patch in such a case.

kubemarine/patches/strong_ciphers_patch.py
```diff
@@ -0,0 +1,52 @@
...
+
+            # Dump the modified YAML back to a string
+            updated_config = yaml.dump(apiserver_config)
+
+            # Write the updated configuration to the file
+            member_node.put(StringIO(updated_config), apiserver_file, backup=True, sudo=True)
```
ilia1243, line 36: If I correctly understand (but not 100% sure), kubelet will restart the kube-apiserver automatically. I suggest to add code that checks that it is restarted successfully, and probably even force restart it to be sure. This may be necessary for further patches to use kubectl and do not fail accidentally if it is not available.

kubemarine/patches/strong_ciphers_patch.py
nikhil1697: Added a logic to check if it already has the --tls-cipher-suite before applying patch

nikhil1697: Okay referring configuration from initial inventory file now

kubemarine/patches/strong_ciphers_patch.py
nikhil1697: Yes added the code to verify and restart the kube-apiserver pod

kubemarine/patches/strong_ciphers_patch.py
```diff
@@ -0,0 +1,113 @@
...
+        cluster = res.cluster()
+
+        # Access the raw inventory directly
+        raw_inventory = res.raw_inventory()
+
+        # Check for probable configuration in the initial inventory
+        probable_ciphers_initial = raw_inventory.get("services", {}).get("kubeadm", {}).get("apiServer", {}).get("extraArgs", {}).get("tls-cipher-suites")
+
+        # Compare with desired ciphers
+        desired_ciphers = "TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384,TLS_CHACHA20_POLY1305_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,TLS_RSA_WITH_3DES_EDE_CBC_SHA,TLS_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_256_CBC_SHA,TLS_RSA_WITH_AES_256_GCM_SHA384"
```
ilia1243, line 26: Why this differs from defaults.yaml?

kubemarine/patches/strong_ciphers_patch.py
```diff
@@ -0,0 +1,113 @@
...
+        if probable_ciphers_initial == desired_ciphers:
```
ilia1243, line 27: Custom ciphers can present and be not equal to the default ones. In this case we have to preserve custom ciphers.

kubemarine/patches/strong_ciphers_patch.py
```diff
@@ -0,0 +1,113 @@
...
+            else:
+                cluster.log.error("API server restart failed.") 
+
+    def verify_apiserver_restart(self, member_node: NodeGroup, res: DynamicResources) -> bool:
+        # Implement logic to check API server restart status
+        cluster = res.cluster()
+        restart_successful = False
+
+        # Get the creation timestamp of the kube-apiserver pod
+        pod_creation_timestamp = member_node.run("sudo kubectl get pod -n kube-system -l component=kube-apiserver -o jsonpath='{.items[0].metadata.creationTimestamp}'")
```
ilia1243, line 74: I tried to run the patch on all-in-one scheme and got

```
Unable to connect to the server: EOF
error executing jsonpath "{.items[0].metadata.creationTimestamp}": Error executing template: array index out of bounds: index 0, length 0. Printing more information for debugging the template:
template was:
{.items[0].metadata.creationTimestamp}
object given to jsonpath engine was:
map[string]interface {}{"apiVersion":"v1", "items":[]interface {}{}, "kind":"List", "metadata":map[string]interface {}{"resourceVersion":""}}
```

kubemarine/patches/strong_ciphers_patch.py
```diff
@@ -0,0 +1,113 @@
...
+
+        cluster.log.debug(f"Original pod creation timestamp: {pod_creation_timestamp}")
+
+        # Optionally, force restart the kube-apiserver pod
+        member_node.run("sudo kubectl delete pod -n kube-system -l component=kube-apiserver")
```
ilia1243, line 79: This way does not restart the container of static pod. You can check that still old container is running using `crictl ps`.

kubemarine/patches/strong_ciphers_patch.py
```diff
@@ -0,0 +1,113 @@
...
+
+            # Call verify_apiserver_restart to check API server restart
+            restart_successful = self.verify_apiserver_restart(member_node, res)
+
+            if restart_successful:
+                cluster.log.info("API server restart succeeded.")
+            else:
+                cluster.log.error("API server restart failed.") 
+
+    def verify_apiserver_restart(self, member_node: NodeGroup, res: DynamicResources) -> bool:
```
ilia1243, line 68: There are a few bugs and gaps below that would not arise if you followed one of existing approaches mentioned in https://github.com/Netcracker/KubeMarine/pull/581#discussion_r1458574834

kubemarine/patches/strong_ciphers_patch.py
ilia1243, line 74: .items[0] - why checks only first control plane?

kubemarine/patches/strong_ciphers_patch.py
ilia1243: Why resolved? I checked, and the new arg is not persisted.

kubemarine/patches/strong_ciphers_patch.py
nikhil1697: Some ciphers were missed in the defaults.yaml, updated it now the ciphers are same

kubemarine/patches/strong_ciphers_patch.py
nikhil1697: Okay now to resolve this in the updated code the defaults ciphers patch will be skipped in case custom ciphers are already present.

kubemarine/patches/strong_ciphers_patch.py
nikhil1697: Followed the existing approach as specified in the comment

kubemarine/patches/strong_ciphers_patch.py
nikhil1697: Removed checking on the first control plane, now all 3 pod status will be visible.

nikhil1697: Now it works on all-in-one scheme aswell

ilia1243:
Pretty good this time. Still, I suggest to postpone this PR. Coincidentally, new cluster reconfigure procedure is coming #592 that will allow to reconfigure kube-apiserver in one call `kubemarine.kubernetes.component.reconfigure_components(kubernetes_nodes, ['kube-apiserver'])`. You may already test it, but please do not rebase, that PR is going to be forced updated.

ilia1243 (approved):

PR #589, opened 2024-01-23
Allow plugins without installation procedures

theboringstuff:
### Description
Plugins which do not have `installation.procedures` cause upgrade failures
```
File "/usr/local/lib/python3.12/site-packages/kubemarine/thirdparties.py", line 418, in <genexpr>
     for plugin_procedure in plugin_configs['installation']['procedures']
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
 KeyError: 'procedures'
```

### Solution
Set keys `installation` and `installation.procedures` as optional, allowing empty values.

### Test Cases

**TestCase 1**

Steps:

1. Add a custom plugins which does not have `installation.procedures` sections, run `upgrade`

Results:

| Before | After |
| ------ | ------ |
| Upgrade fails on thirdparties  | Upgrade continues |

### Checklist
- [ ] Integration CI passed

### DIFF
--- a/kubemarine/thirdparties.py
+++ b/kubemarine/thirdparties.py
@@ -415,7 +415,7 @@ def install_all_thirparties(group: NodeGroup) -> None:
         if cluster.context.get("initial_procedure") in ("install", "upgrade"):
             managing_plugin = next((plugin_name
                                     for plugin_name, plugin_configs in cluster.inventory['plugins'].items()
-                                    for plugin_procedure in plugin_configs['installation']['procedures']
+                                    for plugin_procedure in plugin_configs.get('installation', {}).get('procedures', [])
                                     if plugin_procedure.get('thirdparty') == destination),
                                    None)
 


ilia1243 (approved):

kubemarine/thirdparties.py
```diff
@@ -415,7 +415,7 @@ def install_all_thirparties(group: NodeGroup) -> None:
         if cluster.context.get("initial_procedure") in ("install", "upgrade"):
             managing_plugin = next((plugin_name
                                     for plugin_name, plugin_configs in cluster.inventory['plugins'].items()
-                                    for plugin_procedure in plugin_configs['installation']['procedures']
+                                    for plugin_procedure in plugin_configs.get('installation', {}).get('procedures', [])
```
ilia1243, line 418: From the point of view of Kubemarine, empty installation procedures does not make sense. There maybe even an opposite change, i. e. to introduce restriction of non-empty list to highlight that something is misconfigured / forgotten.

By allowing the empty list we imply that plugins become not something that managed by Kubemarine solely. They become something that may be managed externally, in addition to already existing `values` section.

theboringstuff: > From the point of view of Kubemarine, empty installation procedures does not make sense.

This is true for existing plugins. But **potentially**, KubeMarine may have some **OOB** plugin which does not need installation procedure. Such plugin only need to be "enabled" by `install: true`. And this "enablement" itself will be sufficient for plugin to be used in some KubeMarine task/logic, there is no installation procedure for it. But I agree that for **user-defined/custom** plugins such procedure-less plugins do not make sense and most probably a mistake

ilia1243 (approved):

PR #576, opened 2023-12-14
[CPDEV-97376] Support of Kubernetes v1.29.1

ilia1243:
### Description
* The PR  to support Kubernetes v1.29.1
* Update all third-parties to their latest stable versions
* 
### Solution
* Added programmatic API to reconfigure kube-proxy based on https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-reconfigure/#applying-kube-proxy-configuration-changes
* Added new default for `services.kubeadm_kube-proxy.conntrack.min` property.
* Added fine-grained renew of `super-admin.conf`.

### Test Cases

* Install v1.29.1
* Upgrade from v1.28.x to v1.29.1
* Renew `super-admin.conf`.

### Checklist
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] Integration CI passed
- [x] Unit tests. If Yes list of new/changed tests with brief description
- [x] There is no merge conflicts

#### Unit tests
test_detaults.py - cover enrichment of `services.kubeadm_kube-proxy.conntrack.min` property.

PR #615, opened 2024-03-04
[CPDEV-98165] Fix check_paas services.system.sysctl.config task failure

n549:
### Description
Check_paas's task `services.system.sysctl.config` fails if `services.kubeadm_kube-proxy.conntrack.min`'s custom value is defined in cluster.yaml.

Fixes # (issue)
CPDEV-98165

### Solution
After [this change](https://github.com/Netcracker/KubeMarine/pull/576/files#diff-39319040969c955b67cc5e18dad37e9d99e60833cd002e16df8ea1fde8b0fafaR43) the `net.netfilter.nf_conntrack_max` kernel variable is defined in 2 places of cluster.yaml: in `services.sysctl.net.netfilter.nf_conntrack_max` and in `services.kubeadm-kube-proxy.conntrack.min` and the latter is applied later during the installation process. So eventually `net.netfilter.nf_conntrack_max` value at the cluster node differs from  `services.sysctl.net.netfilter.nf_conntrack_max` value in cluster.yaml.

So 
1. `kubernetes.enrich_inventory` procedure was updated with `services.kubeadm-kube-proxy.conntrack.min` overriding for k8s 1.29+ with  `services.sysctl.net.netfilter.nf_conntrack_max` and for older versions with empty string.
2. `kubemarine.core.defaults.manage_primitive_values` task is moved after `kubemarine.kubernetes.enrich_inventory` in the DEFAULT_ENRICHMENT_FNS list so the overridden value of `services.kubeadm-kube-proxy.conntrack.min` is also update.
3. Doc update: `kubeadm_kube-proxy` part of Installation.md is adjusted accordingly.
4. Doc update: `kubeadm_patches` part of Installation.md is updated to clarify work with extraArgs of control-plane pods (no relations to CPDEV-98165). 

### How to apply
-
### Test Cases
1. Deploy a cluster with custom `services.kubeadm-kube-proxy.conntrack.min` value.
ER:  In the `kube-proxy` configmap the value of `conntrack.min` is the same as `services.sysctl.net.netfilter.nf_conntrack_max` and the same as `net.netfilter.nf_conntrack_max` in /etc/sysctl.d/98-kubemarine-sysctl.conf.

2. Run check_paas against the cluster.
ER: all checks are successful.

### Checklist
- [ ] Integration CI passed

#### Unit tests


kubemarine/kubernetes/__init__.py
```diff
@@ -241,6 +241,13 @@ def enrich_inventory(inventory: dict, cluster: KubernetesCluster) -> dict:
     preflight_errors.extend(default_preflight_errors)
     inventory["services"]["kubeadm_flags"]["ignorePreflightErrors"] = ",".join(set(preflight_errors))
 
+    # override kubeadm_kube-proxy.conntrack.min with sysctl.net.netfilter.nf_conntrack_max since they define the same kernel variable
+    version_key = utils.version_key(get_initial_kubernetes_version(inventory))
+    if version_key >= (1, 29, 0):
+        inventory["services"]["kubeadm_kube-proxy"]["conntrack"]["min"] = inventory["services"]["sysctl"]["net.netfilter.nf_conntrack_max"]
+    else:
+        inventory["services"]["kubeadm_kube-proxy"]["conntrack"]["min"] = ''
```
ilia1243, line 249: More simpler will be to `del` the value, and do not change the order of enrichment functions.

`manage_primitive_values` is designed to be just after `compile_inventory`. Otherwise, it is too easy to introduce some code between `compile_inventory` and `manage_primitive_values` that will rely on, for example, plugins.install, and will not take into account that it may be a string value.

Also, I suggest to remove corresponding jinja template from defaults.yaml and one related section from `manage_primitive_values` as no longer necessary.

n549: Done.

ilia1243 (approved):

PR #617, opened 2024-03-05
Revise configuration of pylint and fix/suppress messages

ilia1243:
### Description
* Pylint is run in CI but is useless due to a bulk of messages.
* It is possible to check some typical problems using pylint and make CI fail if there are any.

### Solution
* Added pylint configuration in pyproject.yaml.
* Fail CI action if any message is printed.
* Disable some frequent unimportant messages.

### Checklist
- [x] Integration CI passed

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,8 +44,12 @@ jobs:
           python-version: "3.9"
       # Install pylint and kubemarine with all dependencies except ansible.
       # Then uninstall only kubemarine to avoid ambiguity and to surely run pylint on sources.
-      - run: pip install pylint . && pip uninstall -y kubemarine
-      - run: pylint kubemarine scripts test --disable fixme || exit 0
+      - run: pip install .[pylint] && pip install -r requirements-pyinstaller.txt && pip uninstall -y kubemarine
+      - run: |
+          if ! pylint kubemarine scripts ; then
+            echo -e "\033[91mCheck [tool.pylint.main] in pyproject.toml for how to fix pylint check\033[0m"
+            exit 1
+          fi
   radon:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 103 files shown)


pyproject.toml
```diff
@@ -95,6 +98,82 @@ push = true
...
+    "unspecified-encoding",
+    # List of other warnings that are considered more error prone
+    "arguments-differ",
+    "arguments-renamed",
+    "cell-var-from-loop",
+    "implicit-str-concat",
+    "undefined-loop-variable",
+    # List of other non-critical warnings that rarely appear
+    "useless-parent-delegation",
+]
```
ilia1243, line 133: Currently, pylint output is clear.
`fail-under = 9.9` allows some messages, but the CI job will eventually fail after the code is spoiled.
Alternatives are:
* Fail if any message is printed. This will make development more complex.
* Fail for any warning. Less complex, but some warnings are still quite unimportant, e.g. `useless-else-on-loop`.
* Do not use fail-under. The cons is as long as the output rise, it will become harder to find messages that really caused the failure, e.g. `consider-using-sys-exit`.

.github/workflows/tests.yaml
```diff
@@ -44,8 +44,12 @@ jobs:
           python-version: "3.9"
       # Install pylint and kubemarine with all dependencies except ansible.
       # Then uninstall only kubemarine to avoid ambiguity and to surely run pylint on sources.
-      - run: pip install pylint . && pip uninstall -y kubemarine
-      - run: pylint kubemarine scripts test --disable fixme || exit 0
+      - run: pip install .[pylint] && pip install -r requirements-pyinstaller.txt && pip uninstall -y kubemarine
+      - run: |
+          if ! pylint kubemarine scripts ; then
```
Imadzuma, line 49: Are you sure, that we should not check `test` package? Although this is a dev code, it's also read/rewritten/reviewed by developers, for this reason I think, that using linters is a good practice for tests too

ilia1243: It introduces ~300 more messages that should be fixed now, and maintained later. Ok, I'll fix them, but what about mypy in this case, do you wish to run it on tests too?

Imadzuma (approved):

pyproject.toml
ilia1243: Decided to fail if any message is printed.

PR #659, opened 2024-05-13
[CPDEV-101682] Support single control plane cluster

ilia1243:
### Description
* If no nodes with `control-plane` role are provided, Kubemarine fails with internal error.
* At least one control-plane node is necessary even for connection-only enrichment
    * choose default node in `do` procedure
    * preserve information about the procedure run.
* In contrast, it is potentially possible to install the cluster with the only `control-plane` node for dev aims.
    * In particular, this allows to rebuild the only worker node using remove & add node while temporarily persisting only control planes.

### Solution
* Raise exception if no control planes are specified in the inventory.
* Allow clusters without workers.

### Test Cases

**TestCase 1**

Rebuild the only worker node using remove & add node.

Steps:

1. Install the cluster with 1 worker and 1 control-plane.
2. Remove the only worker.

Results:

| Before | After |
| ------ | ------ |
| KME0004 | Node is removed successfully |

3. Add the worker.

ER: Node is added successfully.

**TestCase 2**

Install single control plane cluster.

Steps:

1. Configure inventory
   ```yaml
   nodes:
   - name: control-plane
     internal_address: <address>
     roles: [control-plane]
   services:
     coredns:
       deployment:
         spec:
           template:
             spec:
               nodeSelector:
                 node-role.kubernetes.io/worker: null
   plugins:
     nginx-ingress-controller:
       install: false
   ```
2. Install the cluster.

Results:

| Before | After |
| ------ | ------ |
| KME0004 | Cluster is installed successfully |

3. Upgrade the cluster.

ER: Cluster is upgraded successfully.

**TestCase 3**

Try to install cluster without control planes.

Steps:

1. Try to install cluster without control planes.

Results:

| Before | After |
| ------ | ------ |
| Internal error | KME0004  |

### Checklist
- [x] Integration CI passed

#### Unit tests
test_defaults.yaml - cover new functionality and validation.

### DIFF
--- a/kubemarine/kubernetes/__init__.py
+++ b/kubemarine/kubernetes/__init__.py
@@ -108,7 +108,11 @@ def enrich_reconfigure_inventory(cluster: KubernetesCluster) -> None:
 
 @enrichment(EnrichmentStage.ALL)
 def verify_roles(cluster: KubernetesCluster) -> None:
-    if cluster.make_group_from_roles(['control-plane']).is_empty():
+    control_plane_roles = ['control-plane']
+    if cluster.context['initial_procedure'] == 'do':
+        control_plane_roles = ['control-plane', 'master']
+
+    if cluster.make_group_from_roles(control_plane_roles).is_empty():
         raise KME("KME0004")
 
 
--- a/kubemarine/kubernetes/__init__.py
+++ b/kubemarine/kubernetes/__init__.py
@@ -106,6 +106,12 @@ def enrich_reconfigure_inventory(cluster: KubernetesCluster) -> None:
         default_merger.merge(cluster.inventory.setdefault('services', {}), utils.deepcopy_yaml(kubeadm_sections))
 
 
+@enrichment(EnrichmentStage.ALL)
+def verify_roles(cluster: KubernetesCluster) -> None:
+    if cluster.make_group_from_roles(['control-plane']).is_empty():
+        raise KME("KME0004")
+
+
 @enrichment(EnrichmentStage.FULL)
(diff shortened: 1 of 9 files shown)


kubemarine/kubernetes/__init__.py
```diff
@@ -106,6 +106,12 @@ def enrich_reconfigure_inventory(cluster: KubernetesCluster) -> None:
         default_merger.merge(cluster.inventory.setdefault('services', {}), utils.deepcopy_yaml(kubeadm_sections))
 
 
+@enrichment(EnrichmentStage.ALL)
+def verify_roles(cluster: KubernetesCluster) -> None:
+    if cluster.make_group_from_roles(['control-plane']).is_empty():
```
Imadzuma, line 111: I'm not sure about creating clusters without worker node, because such cluster looks useless for dev case too. For example such cluster can't be installed by default via kubemarine (it fails on calico plugin installation because of pending coredns nodes). 
I can not imagine, in such case only control-plane should be needed. For dev cases it's better to use simple all-in-one cluster.

ilia1243: Thank you for feedback.
`verify_roles` checks presence of at least one CP, so no problem here. I suppose you mean potentially adding check of at least one worker.
Single CP cluster might still be installed, see **TestCase 2**, but I agree that it is pretty much useless for dev aims.
I see more useful for dev aims **TestCase 1** which may temporarily leave cluster on one CP.

OleksandrIaresko (approved):

PR #690, opened 2024-09-05
[CPREQ-10003] - deprecated kuber version

Imadzuma:
### Description
Kubernetes versions are excluded from support in kubemarine without any warnings (except mentions in release notes).
But user should plan your upgrade to a new version in advance, so should know when the version he is using becomes a deprecated version and will soon be disabled.

### Solution
* Added version deprecation support in compatibility map:
  * Reused `supported` field to mark version as deprecated;
  * Fixed logs after `install`/`upgrade` procedures, warn if version is deprecated;
  * Added new check-paas task, that checks, if version is deprecated;
* Kubernetes `1.26` is marked as deprecated;
* Added validation check for sync tool
  * If default version does not exist in compatibility map (should be removed), sync tool fails with exception;
* Other fixes:
  * Fixed docker build issue: fixed tags for base images (specified patch version);
     * Root cause was in new `python:3.12.5`, where `setuptools` was excluded by default; more info https://github.com/docker-library/python/issues/952
  * Fixed used python versions in tests;

### Test Cases

**TestCase 1**

Test Configuration:

- Hardware: 
- OS: 
- Inventory: kubernetes 1.26

Steps:

1. Run `kubemarine install`;
2. Run `kubemarine check_paas`;

Results:

| Before | After |
| ------ | ------ |
| No warnings | Warning in the end, if version is deprecated |
| Not applicatable | New task, that verifies kubernetes version, warns, that version is deprecated |

**TestCase 2**

Steps:

1. Try to remove 1.28 kubernenes version using `sync` tool (default version is `1.28.12`);

Results:

| Before | After |
| ------ | ------ |
| Procedure is successfull | Procedure fails with an exception |

### Checklist
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1
 
 # Build ipip_check binary
-FROM golang:1.22 AS go-build
+FROM golang:1.22.5 AS go-build
 
 WORKDIR /opt
 
@@ -11,7 +11,7 @@ RUN go mod download && \
     GOOS=linux CGO_ENABLED=1 go build -ldflags="-linkmode external -extldflags='-static'" -o ipip_check -buildvcs=false && \
     gzip ipip_check
 
-FROM python:3.12-slim-bullseye AS python-build
+FROM python:3.12.4-slim-bullseye AS python-build
 
 ARG BUILD_TYPE
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1
 
 # Build ipip_check binary
-FROM golang:1.22.5 AS go-build
+FROM golang:1.22 AS go-build
 
 WORKDIR /opt
 
@@ -11,7 +11,7 @@ RUN go mod download && \
     GOOS=linux CGO_ENABLED=1 go build -ldflags="-linkmode external -extldflags='-static'" -o ipip_check -buildvcs=false && \
     gzip ipip_check
 
-FROM python:3.12.4-slim-bullseye AS python-build
+FROM python:3.12-slim-bullseye AS python-build
 
 ARG BUILD_TYPE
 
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN go mod download && \
     GOOS=linux CGO_ENABLED=1 go build -ldflags="-linkmode external -extldflags='-static'" -o ipip_check -buildvcs=false && \
     gzip ipip_check
 
-FROM python:3.12-slim-bullseye AS python-build
+FROM python:3.12.4-slim-bullseye AS python-build
```
edmorley, line 14: H! Instead of downgrading to Python 3.12.4, I would recommend adding `setuptools` to the pip install command here (ie installing both `build` and `setuptools`):
https://github.com/Netcracker/KubeMarine/blob/1b3e612e1eff7fbdcace47bbfc58ac8636ad8c0b/Dockerfile#L28

And then longer term I would report a bug with whatever dependency is using setuptools but hasn't correctly listed it in its requirements (or similar).

Imadzuma: Hi, @edmorley!
Initially, I wanted to fix the patch version to avoid such issues in the future, because they can break pipelines during the development of other features (like now). Updating the version is our regular task, so we would not get stuck on `3.12.4`.
 But after a private discussion with @koryaga , it was decided to leave everything as is, and resolve only this particular issue.

kubemarine/procedures/check_paas.py
```diff
@@ -1565,6 +1565,18 @@ def kubernetes_admission_status(cluster: KubernetesCluster) -> None:
         cluster.log.debug(kube_admission_status)
         tc.success(results='enabled')
 
+def verify_kubernetes_version(cluster: KubernetesCluster) -> None:
```
koryaga, line 1568: Need to update https://github.com/Netcracker/KubeMarine/blob/main/documentation/Kubecheck.md#paas-procedure respectively

koryaga:
@OleksandrIaresko  pls proceed with QA

koryaga:
@disa1217  pls validate respective output

disa1217:
@koryaga , reviewed and made a few changes. please check.

disa1217 (approved):

vlar0816 (approved):

kubemarine/procedures/check_paas.py
Imadzuma: done

PR #720, opened 2025-02-19
Troubleshooting guide update

alexarefev:
### Description
* Common network issues should be described in documentation 

### Solution
* Documentation update

### Checklist
- [x] Integration CI passed


documentation/Troubleshooting.md
```diff
@@ -2012,4 +2013,23 @@ Failed to start OpenSSH server daemon.
...
 - Test updates in a staging environment to catch compatibility issues before deployment.
 
-**Note**: Not applicable.
\ No newline at end of file
+**Note**: Not applicable.
+
+## Packets loss during the transmission between nodes
+
+### Description
+Packets are lost during the transmission between nodes that are located in defferent subnets. It appears in retries of TCP sessions or inability to get the UDP packets in case of high network load. The root cause is in theIaaS level routers performance. Basically, routing works slower than switching.
```
pranavcracker, line 2021:
```suggestion
Packets are lost during the transmission between nodes that are located in different subnets. It appears in retries of TCP sessions or inability to get the UDP packets in case of high network load. The root cause is in the IaaS level routers performance. Basically, routing works slower than switching.
```

pranavcracker (changes requested):

pranavcracker (approved):

DmitriiRabenok (approved):

PR #732, opened 2025-03-20
Installation guide update. Centos7.5 support removal.

iagusev:
### Description

Documentation update.
Centos 7.5, RHEL 7.5 and Oracle Linux 7.5 has been removed from prerequisites and v1.32 table.

### Checklist
- [x] Integration CI passed


pranavcracker:
I think for removing support of CentOS 7.5, only removal from installation.md would not be enough.
We have remove it from form configurations 
e.g. https://github.com/Netcracker/KubeMarine/blob/main/kubemarine/resources/configurations/globals.yaml#L303

documentation/Installation.md
```diff
@@ -163,9 +163,9 @@ For cluster machines, ensure the following requirements are met:
 
 * The following distributives and versions are supported:
 
-  * Centos 7.5+, 8.4, 9
```
alexarefev, line 166: Do we need to exclude only CentOS 7.5 version or all of the versions from 7.5 and higher?

DmitriiRabenok:
Will be fully covered in https://github.com/Netcracker/KubeMarine/pull/749

PR #736, opened 2025-04-10
Fix Critical & High Vulnerabilities in Kubemraine

pranavcracker:
### Description
* When we scan kubermarine v0.36.0 image we found below packages are being vulnerable - 

1. golang.org/x/crypto
2. github.com/docker/docker
3. github.com/golang/go
4. zlib1g
5. golang.org/x/crypto
6. golang.org/x/oauth2
7. ansible-core
8. perl-base
9. pip

### Solution
* Update to latest version which has fixes for the vulnerabilities  

Below is the status for vulnerability fixes - 
| Vulnerable Package       | Where It Is Present                        | Fixed (Yes/No) | Remarks                                                              |
|--------------------------|--------------------------------------------|----------------|----------------------------------------------------------------------|
| golang.org/x/crypto      | Helm                                       | Yes            | Updated Helm to v3.17.3                                              |
| github.com/docker/docker | Helm                                       | Yes            | Updated Helm to v3.17.3                                              |
| github.com/golang/go     | Helm                                       | Yes            | Updated Helm to v3.17.3                                              |
| golang.org/x/oauth2      | Helm                                       | Yes            | Updated Helm to v3.17.3                                              |
| ansible-core             | Ansible                            | Yes            | Updated ansible version to 11.4                                      |
| perl-base                | Debian package in python:3.12-slim-bookworm| No             | No fixed version available even in latest Python/Debian release      |
| pip                      | Debian package in python:3.12-slim-bookworm| No             | No fixed version available even in latest Python/Debian release      |
| zlib1g                   | Debian package in python:3.12-slim-bookworm| No             | No fixed version available even in latest Python/Debian release      |


kubemarine/resources/scripts/source/ipip_check/go.mod
```diff
@@ -1,10 +1,42 @@
 module ipip_check
 
-go 1.22.4
+go 1.23.6
```
theboringstuff, line 3: Why do we update to `1.24` in some places, but in other to `1.23` only?

kubemarine/resources/scripts/source/ipip_check/ipip_check.go
```diff
@@ -1,47 +1,53 @@
...
+	"errors"
+	"flag"
 	"fmt"
 	"net"
 	"os"
 	"time"
-	"flag"
-	"errors"
 
+	_ "github.com/docker/docker/client"
```
theboringstuff, line 32: Please add comment why we import these packages here, because it may not be clear after some time

pyproject.toml
```diff
@@ -54,7 +54,7 @@ dependencies = [
 requires-python = ">=3.8"
 
 [project.optional-dependencies]
-ansible = ["ansible==7.7.*"]
+ansible = ["ansible==11.4.*"]
```
theboringstuff, line 57: This looks like it could be a breaking change for some users. As I understand, this ansible dependency is used to execute user-provided ansible plugins. Thus, after such update, it is possible that user ansible plugins may break. We probably need to communicate this change to users, so that they are prepared to take action on their side

kubemarine/resources/scripts/source/ipip_check/ipip_check.go
pranavcracker: comment added.

kubemarine/resources/scripts/source/ipip_check/go.mod
pranavcracker: updated to 1.24 at all places

pyproject.toml
pranavcracker: created a short doc for highlighting braking changes. we will add link to it in our Release Notes.

kubemarine/resources/scripts/source/ipip_check/ipip_check.go
```diff
@@ -1,47 +1,54 @@
...
+	"flag"
 	"fmt"
 	"net"
 	"os"
 	"time"
-	"flag"
-	"errors"
 
+	// Some packages have indirect dependancy which require their version to be updated. For such packages we have blank imports here as _"package_name"
+	_ "github.com/docker/docker/client"
```
theboringstuff, line 33: This is also not used in ipip check

kubemarine/resources/scripts/source/ipip_check/ipip_check.go
```diff
@@ -1,47 +1,54 @@
...
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
+	_ "github.com/open-policy-agent/opa/v1"
+	_ "golang.org/x/crypto/ssh"
```
theboringstuff, line 37: This also does not seem to be used in ipip check

theboringstuff (approved):

PR #739, opened 2025-04-23
Support unauthenticated OIDC discovery endpoint

theboringstuff:
### Description
Some applications do not support authenticated OIDC discovery endpoints, but Kubernetes by default enables authentication https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery

Need to disable this authentication by default, but allow to enable it if needed

### Solution
* added new boolean parameter `rbac.authenticated-issuer-discovery` which is set to `false` by default (disable authentication)
* `deploy.accounts` install task was extended to apply/delete CRB which allows unauthenticated users to access OIDC discovery endpoints
* added patch to apply new default on existing clusters
* added documentation

### How to apply
* `migrate_kubemarine` with patch `configure_sa_issuer_discovery`

### Test Cases

**TestCase 1**

New KubeMarine version migration should enable unauthenticated access to OIDC discovery endpoint

Steps:
1. Install cluster with previous KubeMarine version
2. Access OIDC discovery endpoint, e.g. `curl https://${CLUSTER_IP}:6443/.well-known/openid-configuration/ -k`
3. You should see 403 Forbidden error
4. Take new KubeMarine version and run `migrate_kubemarine` (do not modify `cluster.yaml`)
5. OIDC discovery endpoint should become available by default

**TestCase 2**

New KubeMarine version clean installation should allow unauthenticated access to OIDC discovery endpoint.
It should be possible to enable authentication afterwards.

Steps:
1. Install cluster with new KubeMarine version
2. Access OIDC discovery endpoint, e.g. `curl https://${CLUSTER_IP}:6443/.well-known/openid-configuration/ -k`
3. You should not see any error, endpoint should be available
4. Change cluster.yaml, set `rbac.authenticated-issuer-discovery: true`
5. Run install procedure with `deploy.accounts` task
6. OIDC discovery endpoint should become unavailable returning 403 Forbidden

### Checklist
- [x] Integration CI passed

### DIFF
--- a/kubemarine/kubernetes_accounts.py
+++ b/kubemarine/kubernetes_accounts.py
@@ -62,7 +62,8 @@ def handle_authenticated_sa_issuer_discovery(cluster: KubernetesCluster) -> None
     To disable authentication, we apply CRB which allows unauthenticated access.
     To enable authentication, we delete above CRB (if it is present).
 
-    For more information on the cluster role: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery
+    For more information on the cluster role: 
+    https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery
     """
     mode = "unauthenticated"
     kubectl_cmd = "apply"
--- a/kubemarine/kubernetes_accounts.py
+++ b/kubemarine/kubernetes_accounts.py
@@ -58,14 +58,9 @@ def enrich_inventory(cluster: KubernetesCluster) -> None:
 def handle_authenticated_sa_issuer_discovery(cluster: KubernetesCluster) -> None:
     """
     This function handles SA issuer discovery endpoint authentication
-    Endpoint URL is: https://<cluster_ip>:6443/.well-known/openid-configuration/
 
     To disable authentication, we apply CRB which allows unauthenticated access.
     To enable authentication, we delete above CRB (if it is present).
-
-    Enable/disable is controlled by "rbac.authenticated-issuer-discovery" flag.
-    By default, we disable authentication. 
-    This is in contrast to native k8s default, where it is enabled by default.
     """
     mode = "unauthenticated"
     kubectl_cmd = "apply"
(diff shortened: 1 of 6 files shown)


kubemarine/kubernetes_accounts.py
```diff
@@ -54,7 +55,49 @@ def enrich_inventory(cluster: KubernetesCluster) -> None:
...
+    For more information on the cluster role: 
+    https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery
+    """
+    mode = "unauthenticated"
+    kubectl_cmd = "apply"
+    if cluster.inventory.get("rbac", {}).get("authenticated-issuer-discovery", False):
+        mode = "authenticated"
+        kubectl_cmd = "delete --ignore-not-found"
+    issuer_discovery_crb = dedent("""\
+        kind: ClusterRoleBinding
```
alexarefev, line 74: Maybe we could put it into `defaults.yaml` like that one: https://github.com/Netcracker/KubeMarine/blob/99f13807ed0f79bc9448c371eeb2951e15466b9b/kubemarine/resources/configurations/defaults.yaml#L868

theboringstuff: We could, if we want this CRB to be customizable by user, but I think we do not need that. Also, the less parameters we expose to the user, the less to support. I suggest to keep this CRB hidden in code, if possible

alexarefev: Today we need to allow unauthenticated access to the one endpoint, tomorrow it will need another one. It's much better to have generalized approach.  And let's user customize that part of configuration.

theboringstuff: For users, we have generalized approach - plugins. Unfortunately, for our usecase plugins do not work in its current state - they do not support "uninstall". They only support "install", or nothing. We need "uninstall" to allow authenticated access. This requires enhancement to plugins, which is out of scope for this feature I believe

DmitriiRabenok (approved):

PR #767, opened 2025-08-28
[CPCAP-760]_Removed centos 7&8 and rhel 7&8

nikhil1697:
### Description
Drop support for RHEL/CentOS 7–8; standardize HAProxy to native haproxy on RHEL 9; update tests and docs

### Solution
1. Removes RHEL 7/8 and CentOS 7/8 from supported OS list.
2. Switches HAProxy service artifacts from legacy SCL (rh-haproxy18-haproxy) to native haproxy on RHEL 9.
3. Aligns code paths, tests, and documentation with the supported OS families: debian and rhel9.

### Checklist
- [ ] There is no breaking changes, or migration patch is provided
- [ ] Integration CI passed

#### Unit tests

### DIFF
--- a/documentation/Installation.md
+++ b/documentation/Installation.md
@@ -165,16 +165,16 @@ For cluster machines, ensure the following requirements are met:
 
 * The following distributives and versions are supported:
 
-  * Centos 7.5+, 8.4, 9
-  * RHEL 7.5+, 8.4, 8.6, 8.7, 8.8, 8.9, 9.2
-  * Oracle Linux 7.5+, 8.4, 9.2
+  * CentOS Stream 9
+  * RHEL 9.2, 9.3, 9.4
+  * Oracle Linux 9.2
   * RockyLinux 8.6, 8.7, 8.8, 9.2, 9.3 ,9.4, 9.5
   * Ubuntu 20.04, 22.04.1, 24.04.1
  
 **Note**: Ubuntu 24.04 is supported only with kubernetes versions starting from v1.29.7 and above.
 
 <!-- #GFCFilterMarkerStart# -->
-The actual information about the supported versions can be found in `compatibility_map.distributives` section of [globals.yaml configuration](../kubemarine/resources/configurations/globals.yaml#L256).
+The actual information about the supported versions can be found in `compatibility_map.distributives` section of [globals.yaml configuration](../kubemarine/resources/configurations/globals.yaml#L299).
 <!-- #GFCFilterMarkerEnd# -->
 
 **Networking**
@@ -244,13 +244,7 @@ For example, specify `conntrack-tools` instead of `conntrack`.
 **Note**:
 
 * You can install a version other than the recommended version, but it is not supported and can cause unpredictable consequences.
-* rh-haproxy18 (build provided by RedHat) is supported only for now.
-
-**Warning**: RHEL 8 does not have Python preinstalled. For `check_iaas` to work correctly, it is required to install Python on the nodes. Execute the following step before the installation procedure.
-* Install `python 3.9` from the standard RHEL repository:
-```
-# dnf install python3.9 -y
-```
+* Note: RHEL 7/8 and CentOS 7/8 are no longer supported.
 
 **Unattended-upgrades** 
 
@@ -1622,7 +1616,7 @@ services:
           name: "Kubernetes"
           enabled: 1
           gpgcheck: 0
-          baseurl: "https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64"
+          baseurl: "https://packages.cloud.google.com/yum/repos/kubernetes-el9-x86_64"
         my_own_repo:
           name: "My repository"
           enabled: 1
@@ -1825,15 +1819,15 @@ The following associations are used by default:
   <tr>
     <td rowspan="4">haproxy</td>
     <td>executable_name</td>
-    <td>/opt/rh/rh-haproxy18/root/usr/sbin/haproxy</td>
+    <td>haproxy</td>
   </tr>
   <tr>
     <td>package_name</td>
-    <td>rh-haproxy18-haproxy-{{k8s-version-specific}}</td>
+    <td>haproxy</td>
   </tr>
   <tr>
     <td>service_name</td>
-    <td>rh-haproxy18-haproxy</td>
+    <td>haproxy</td>
   </tr>
   <tr>
     <td>config_location</td>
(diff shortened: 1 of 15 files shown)


nikhil1697:
recheck

nikhil1697:
I have read the CLA Document and I hereby sign the CLA

kubemarine/demo.py
```diff
@@ -507,7 +507,7 @@ def new_cluster(inventory: dict, procedure_inventory: dict = None, context: dict
 
 def generate_node_context(*,
                           online: bool = True, accessible: bool = True, sudo: str = 'Root',
-                          os_name: str = 'centos', os_version: str = '7.9',
+                          os_name: str = 'rhel', os_version: str = '9.2',
```
pranavcracker, line 510: Instead of changing OS from centos to rhel, can we just update cetnos version to supported one?

test/unit/core/test_cluster.py
```diff
@@ -134,9 +134,9 @@ def test_upgrade_get_redefined_package_associations(self):
     def _nodes_context_one_different_os(self, inventory, host_different_os):
         nodes_context = demo.generate_nodes_context(inventory, os_name='ubuntu', os_version='20.04')
         nodes_context[host_different_os]['os'] = {
-            'name': 'centos',
-            'family': 'rhel',
-            'version': '7.9'
+            'name': 'rhel',
```
pranavcracker, line 137: can we just update cetnos version to supported one?

nikhil1697: In the modified codebase CentOS 9 remains supported, but i consolidated all RHEL‑family derivatives (CentOS, Rocky Linux, Oracle Linux) under a single rhel9 OS family. This was done because CentOS 9 and RHEL 9 are binary compatible and share the same kernel and package ecosystem, 
But if we dont want to make changes and keep as it was before it can be reverted back to centos itself

kubemarine/packages.py
```diff
@@ -481,7 +486,8 @@ def disable_unattended_upgrade(group: NodeGroup) -> None:
 
 
 def get_associations_os_family_keys() -> Set[str]:
-    return {'debian', 'rhel', 'rhel8', 'rhel9'}
+    # Supported OS families after dropping RHEL/CentOS 7/8
+    return {'debian', 'rhel9'}
```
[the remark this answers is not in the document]
nikhil1697: Yes it does, will remove the rhel 7 & 8 entries from rocky and Ol aswell

alexarefev (approved):

documentation/Installation.md
```diff
@@ -244,13 +244,7 @@ For example, specify `conntrack-tools` instead of `conntrack`.
...
 
 * You can install a version other than the recommended version, but it is not supported and can cause unpredictable consequences.
-* rh-haproxy18 (build provided by RedHat) is supported only for now.
-
-**Warning**: RHEL 8 does not have Python preinstalled. For `check_iaas` to work correctly, it is required to install Python on the nodes. Execute the following step before the installation procedure.
-* Install `python 3.9` from the standard RHEL repository:
-```
-# dnf install python3.9 -y
-```
+* Note: RHEL 7/8 and CentOS 7/8 are no longer supported.
```
theboringstuff, line 247: As I understand, OL and Rocky 8 are also not supported. By the way, why we decided to drop support for RHEL, OL and Rocky 8?

kubemarine/packages.py
```diff
@@ -480,8 +480,9 @@ def disable_unattended_upgrade(group: NodeGroup) -> None:
                      sudo=True, backup=True)
 
 
-def get_associations_os_family_keys() -> Set[str]:
-    return {'debian', 'rhel', 'rhel8', 'rhel9'}
+def get_associations_os_family_keys() -> List[str]:
+    # Deterministic order; exclude legacy 'rhel' generic for OS-family loops
+    return ['debian', 'rhel', 'rhel8', 'rhel9']
```
theboringstuff, line 485: Why do we keep `rhel` OS family, if we have no actual versions under this OS family? It seems for me we can remove it. Also, can you please explain what this comment means, I do not get its purpose

test/unit/test_modprobe.py
```diff
@@ -125,7 +125,8 @@ def test_error_invalid_boolean_value_simple_format(self):
             self.new_cluster()
 
     def test_multiple_os_family_merge(self):
-        supported_os_families = list(packages.get_associations_os_family_keys())
+        # Exclude legacy 'rhel' family for merge behavior since no nodes run plain 'rhel'
+        supported_os_families = [f for f in packages.get_associations_os_family_keys() if f != 'rhel']
```
theboringstuff, line 129: It seems that if we remove `rhel` from the list, we do not need to exclude it here

kubemarine/packages.py
nikhil1697: We still have code, schemas, and defaults keyed by 'rhel' even though no supported node resolves to that family anymore.
Examples: [kubemarine/haproxy.py:229](https://github.com/Netcracker/KubeMarine/blob/3f541de5bf73b69fa3cf0429e2c5607e6218c2ab/kubemarine/haproxy.py#L229) (haproxy18 override path for EL7), 
[kubemarine/resources/configurations/defaults.yaml:146](https://github.com/Netcracker/KubeMarine/blob/3f541de5bf73b69fa3cf0429e2c5607e6218c2ab/kubemarine/resources/configurations/defaults.yaml#L146) and 
[kubemarine/resources/configurations/globals.yaml:171](https://github.com/Netcracker/KubeMarine/blob/3f541de5bf73b69fa3cf0429e2c5607e6218c2ab/kubemarine/resources/configurations/globals.yaml#L171) (associations under rhel), 
and the JSON schema allows rhel as a key. Removing 'rhel' from get_associations_os_family_keys() today would require a wider cleanup to avoid breaking inventories and code paths that still reference that key.
And Nodes won’t report os_family='rhel' after dropping EL7, so loops that use this list effectively skip it at runtime (empty subgroup). 

And about the comment it might be confusing,  The intent was to ensure deterministic iteration order. Tests that need to loop only over actual node OS families explicitly filter 'rhel' (see test/unit/test_modprobe.py:128). Will fix the comment

theboringstuff: Users will not be able to use "rhel" OS family anymore, right? So we do not need it, in my understanding. So why do we need to keep it? Is it such significant changes to remove?

nikhil1697: Okay got it, will remove the rhel completly

documentation/Installation.md
```diff
@@ -165,16 +165,16 @@ For cluster machines, ensure the following requirements are met:
```
theboringstuff, line 165: I see rhel 7 still mentioned in several places
* Installation.md
* software_upgrade.yaml (and related files)
* packages.py, yum.py
* in the examples it seems rhel/centos 7 repost are still mentioned

I believe we need to remove those too.

Also, after we change `software_upgrade.yaml` strcuture, I believe we need to dev-test that third-party upgrade procedure still works for packages (e.g. try to upgrade containerd, check that patch is created and that it is applied)

theboringstuff (approved):

PR #775, opened 2025-09-23
Etcd/etcdutl thirdparty support

theboringstuff:
### Description
Starting with k8s 1.34, it uses etcd 3.6, which has removed "restore" functionality from "etcdctl" tool and moved it to a new "etcdutl" tool. We need to start using this new tool.

### Solution
* Added etcdutl thirdparty for all k8s versions
  * We actually require etcdutl only for 1.34, but for simplicity we add it for all k8s versions
  * Etcdutl binary is not available in k8s etcd image, so we have to retrieve it from etcd release archive
  * Etcdutl unpacking is specific, since archive contains many things we do not need
* Restore procedure now uses etcdutl for restoration
* Added support for etcdutl in software_upgrade and migrate_kubemarine
  * Version of etcdutl during software_upgrade is resolved from kubeadm etcd image version
* Added patch for etcdutl installation on migrate_kubemarine
  * For air-gapped envs, migrate_kubemarine patch will require procedure.yaml to define etcdutl source
* Updated docs/tests accordingly

### How to apply
* Run `migrate_kubemarine` to install etcdutl. In air-gapped envs it will be required to provide `procedure.yaml`, e.g.
```
upgrade:
  thirdparties:
    /usr/bin/etcd.tar.gz:
      source: "https://example.com/thirdparty.files/etcd-io/etcd/{{ globals.compatibility_map.software.etcdutl[services.kubeadm.kubernetesVersion].version }}/etcd-{{ globals.compatibility_map.software.etcdutl[services.kubeadm.kubernetesVersion].version }}-linux-amd64.tar.gz"
```

### Test Cases

**Verify migrate_kubemarine installs etcdutl**

Steps:

1. Install cluster with k8s 1.33.0 on KubeMarine v0.40.0
2. Make sure etcdutl binary not present on control-plane nodes (/usr/bin/etcdutl)
3. Take latest KubeMarine version with these changes, run migrage_kubemarine (you may need to provide procedure.yaml with etcdutl source)
4. Make sure etcdutl binary is present on control-plane nodes (/usr/bin/etcdutl)

**Verify restore works**

Steps:

1. Install cluster with any k8s 1.34.0 using latest KubeMarine version with these changes
2. Run backup procedure
3. Run restore procedure to restore from backup, make sure it succeeds

### DIFF
--- a/documentation/Maintenance.md
+++ b/documentation/Maintenance.md
@@ -140,7 +140,7 @@ However, your custom containers may be deleted, and you need to start them manua
 Patches that upgrade thirdparties have the following identifiers:
 * `upgrade_crictl` - It upgrades the `/usr/bin/crictl` third-party, if necessary.
 * `upgrade_calico` - It upgrades the `/usr/bin/calicoctl` third-party as part of the Calico plugin upgrade.
-* `upgrade_etcdutl` - It upgrades the `/usr/bin/etcdutl` third-party as part of the Calico plugin upgrade.
+* `upgrade_etcdutl` - It upgrades the `/usr/bin/etcdutl` third-party, if necessary.
 
 
 If the cluster is located in an isolated environment,
--- a/documentation/Maintenance.md
+++ b/documentation/Maintenance.md
@@ -159,7 +159,7 @@ upgrade:
     /usr/bin/etcd.tar.gz:
       source: "https://example.com/thirdparty.files/etcdutl/etcd/{{
         globals.compatibility_map.software.etcdtul[services.kubeadm.kubernetesVersion].version
-        }}/etcdutl-{{ globals.compatibility_map.software.etcdutl[services.kubeadm.kubernetesVersion].version }}-linux-amd64.tar.gz"
+        }}/etcd-{{ globals.compatibility_map.software.etcdutl[services.kubeadm.kubernetesVersion].version }}-linux-amd64.tar.gz"
 ```
 
 This configuration replaces the configuration contained in the current `cluster.yaml`.
--- a/documentation/Maintenance.md
+++ b/documentation/Maintenance.md
@@ -150,11 +150,16 @@ as shown in the following snippet:
 upgrade:
   thirdparties:
     /usr/bin/calicoctl:
-      source: https://example.com/thirdparty.files/projectcalico/calico/v3.25.1/calicoctl-linux-amd64
+      source: "https://example.com/thirdparty.files/projectcalico/calico/{{
+        plugins.calico.version }}/calicoctl-linux-amd64"
     /usr/bin/crictl.tar.gz:
-      source: https://example.com/thirdparty.files/kubernetes-sigs/cri-tools/v1.27.0/crictl-v1.27.0-linux-amd64.tar.gz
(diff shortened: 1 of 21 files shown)


documentation/Maintenance.md
```diff
@@ -140,6 +140,8 @@ However, your custom containers may be deleted, and you need to start them manua
 Patches that upgrade thirdparties have the following identifiers:
 * `upgrade_crictl` - It upgrades the `/usr/bin/crictl` third-party, if necessary.
 * `upgrade_calico` - It upgrades the `/usr/bin/calicoctl` third-party as part of the Calico plugin upgrade.
+* `upgrade_etcdutl` - It upgrades the `/usr/bin/etcdutl` third-party as part of the Calico plugin upgrade.
```
pranavcracker, line 143: statement needs to updated for etcdutl.

documentation/Troubleshooting.md
```diff
@@ -523,6 +524,12 @@ Since the command is run from a container, this imposes certain restrictions. Fo
 * `/var/lib/etcd`:`/var/lib/etcd`
 * `/etc/kubernetes/pki`:`/etc/kubernetes/pki`
 
+## Etcdutl binary
+
+Starting with etcd v3.6, in addition to `etcdctl` script, there is a new `etcdutl` binary, which took some responsibility from `etcdctl`. This binary is installed during the `prepare.thirdparties` installation task on all control-planes and requires root privileges.
+
+To find out all the available `etcdutl` options and features, use the original ETCD documentation.
```
pranavcracker, line 531: will it be a good idea to provide link to original ETCD documentation here?

documentation/Maintenance.md
theboringstuff: fixed

documentation/Troubleshooting.md
theboringstuff: added links to some example etcdutl docs

scripts/thirdparties/src/software/download.py
nikhil1697, line 1: If the curl operation fails, it would leave TEMP_FILE in an unknown state and concurrent calls could overwrite TEMP_FILE before the rename(race condition),
And can we add a validation for download completed successfully

scripts/thirdparties/src/software/download.py
```diff
@@ -0,0 +1,19 @@
...
+    target_file = os.path.join(SYNC_CACHE, filename)
+    if os.path.exists(target_file):
+        return target_file
+
+    source = thirdparties.get_default_thirdparty_source(destination, version, in_public=True)
+
+    print(f"Downloading thirdparty {destination} of version {version} from {source}")
+    curl(source, TEMP_FILE)
+    os.rename(TEMP_FILE, target_file)
+
```
nikhil1697, line 18: If the curl operation fails, it would leave TEMP_FILE in an unknown state and concurrent calls could overwrite TEMP_FILE before the rename(race condition),
And can we add a validation for download completed successfully

pranavcracker (approved):

DmitriiRabenok (approved):

PR #806, opened 2026-04-27
Add openstack-cinder-csi plugin support

theboringstuff:
### Description
It is required to add openstack-cinder-csi plugin support in KubeMarine

### Solution
* Used official chart https://github.com/kubernetes/cloud-provider-openstack/tree/master/charts/cinder-csi-plugin
* Added support for latest chart version v2.35.0, it is default one
* Added support for old chart version v2.2.0, which could be optionally used for old OpenStacks
* Added documentation about chart usage

### Checklist
- [x] Integration CI passed


documentation/Installation.md
```diff
@@ -4523,6 +4524,74 @@ plugins:
...
+            name: csi-sc-cinderplugin
+          provisioner: cinder.csi.openstack.org
+          reclaimPolicy: Delete
+          volumeBindingMode: Immediate
+          allowVolumeExpansion: true
+          parameters:
+            type: $YOUR_OPENSTACK_VOLUME_TYPE_HERE
+```
+
+**Note:** Above configuration assumes that nodes have OpenStack credentials stored on all nodes under `/etc/config/cloud.conf` file. You need to put this file on nodes during nodes provisioning.
```
alexarefev: To store cloud.conf in the `Secret` is more convenient. It's a single point of truth, there is no additional manipulation on IaaS level.

theboringstuff: It is possible to use in Secret, KubeMarine and chart support that. The problem with Secret is that we currently do not have secure way to pass it to KubeMarine (especially jenkins job), we do not want to provide it plain-text. So for now the most secure way seem to be using file on nodes (confirmed this with SaaS team, they will be able to put it)

kubemarine/resources/configurations/defaults.yaml
```diff
@@ -900,6 +900,31 @@ plugins:
         cpu: 200m
         memory: 256Mi
 
+  openstack-cinder-csi:
+    version: '{{ globals.compatibility_map.software["openstack-cinder-csi"][services.kubeadm.kubernetesVersion].version }}'
+    install: false
+    namespace: "kube-system"
```
pranavcracker: I think we should deploy the cinder-csi plugin in a dedicated namespace instead of kube-system, unless there’s a strict dependency on it. This also aligns with security hardening best practices (e.g., CIS Kubernetes Benchmark) around namespace-level isolation and least privilege.

theboringstuff: I do not think there is any dependency, it is just usual namespace for it, e.g. documented here https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#using-the-helm-chart

But yes, we can probably use dedicated namespace by default, will change

theboringstuff: Changed, now it uses dedicated namespace `openstack-cinder-csi` by default

pranavcracker (approved):

PR #824, opened 2026-08-03
[CPCAP-12251] feature: fsmount

alexarefev:
### Description
* The functionality to mount some storage before the k8s installation is needed
* The ZRAM storage for /var/log/pods must be add to configuration, but disabled by default

### Solution
* Add new task `fsmount` into preparation section with the following description:
```
services:
  fsmount:
  - name: fs1
    size: 1G
    type: ext4
    device: /dev/sdd1
    path: /mnt
    template:
      source: templates/plugins/fsmount-setup.service.j2
      destination: /etc/systemd/system/fsmount-setup.service
    preparation_scripts:
      - resources/scripts/upgrade_kernel.sh
      - resources/scripts/fsprep.sh
	groups: [control-plane, workers]
	nodes: []
```
`name` - name of the item just for indication
`enabled` - defines if the item must be processed
`device` - the device file
`path` - the mount point
`size` - the size of the disk, it's needed for example for tmpfs (optional)
`template.source` - the path to systemctl unit template that should be rendered by the options above
`template.destination` - the path where the systemctl unit must be put on the cluster node
`preparation_script` - the script that must be finished successfully to initiate the filesystem creation (optional); basically it install necessary packages, modules and so no to run systemctl unit successfully; if it finishes with error fsmount task for that item must be skipped
`groups` - the list of the nodes roles where the task must be applied (optional)
`nodes` - the list of the nodes where the task must be applied (optional)
`groups` and `nodes` must be merged before the run; if both groups and nodes are empty - the task is skipped
* Add template, script for ZRAM storage
* Add `check_paas` task
* Update documentation

### Test Cases

**TestCase 1**

Check if the installation procedure is working

Test Configuration:

- Hardware: 2CPU/4GB
- OS: Ubuntu 24.04
- Inventory: Allinone

Steps:

1. Run `kubemarine install` with the following settings in `cluster.yaml`:
```
services:
  kubeadm_kubelet:
    containerLogMaxSize: 5Mi
    containerLogMaxFiles: 2
  fsmount:
  - name: zram
    size: 1G
    type: ext4
    device: /dev/zram0
    path: /var/log/pods
    template:
      source: templates/zram-setup.service.j2
      destination: /etc/systemd/system/zram-setup.service
    preparation_scripts:
      - resources/scripts/upgrade_kernel.sh
      - resources/scripts/zram.sh
    groups: [control-plane, worker]
```

Results:

| Before | After |
| ------ | ------ |
| not applicable | /dev/zram0 is mounted to /var/log/pods; `containerLogMax` options are set in /var/lib/kubelet/config.yaml |

**TestCase 2**

Check if the mount_fs procedure is working

Test Configuration:

- Hardware: 2CPU/4GB
- OS: Ubuntu 24.04
- Inventory: Allinone

Steps:

1. Run `kubemarine reconfigure` on existing cluster with the following settings in `procedure.yaml`:
```
services:
  kubeadm_kubelet:
    containerLogMaxSize: 5Mi
    containerLogMaxFiles: 2
```
2. Run `kubemarine mount_fs` on existing cluster with the following settings in `procedure.yaml`:
```
fsmount:
  - name: zram
    size: 1G
    type: ext4
    device: /dev/zram0
    path: /var/log/pods
    template:
      source: templates/zram-setup.service.j2
      destination: /etc/systemd/system/zram-setup.service
    preparation_scripts:
      - resources/scripts/upgrade_kernel.sh
      - resources/scripts/zram.sh
    groups: [control-plane, worker]
reboot: true
```

Results:

| Before | After |
| ------ | ------ |
| not applicable | /dev/zram0 is mounted to /var/log/pods; `containerLogMax` options are set in /var/lib/kubelet/config.yaml |

### Checklist
- [x] Integration CI passed


disa1217 (approved):
@alexarefev , checked.

docs/public/Installation.md
```diff
@@ -2583,6 +2584,120 @@ The following settings are supported in the extended format:
...
+
+*Can restart a service*: **No**
+
+*Overwrites files*: Yes – the rendered systemd unit file is uploaded to the location defined in `template.destination`. A backup of the previous file is kept.
+
+*OS‑specific*: No
+
+The `services.fsmount` section allows you to define additional filesystems that should be formatted and mounted on cluster nodes using **systemd** unit files. Each list entry describes:
+
+- **name** – an identifier used in logs and dump file names.
```
theboringstuff: It seems this section duplicates below table. Can you please reduce duplication here, so that it is easier to support?

docs/public/Installation.md
```diff
@@ -2583,6 +2584,120 @@ The following settings are supported in the extended format:
...
+    device: /dev/zram0
+    path: /var/log/pods
+    template:
+      source: templates/zram-setup.service.j2
+      destination: /etc/systemd/system/zram-setup.service
+    preparation_script: resources/scripts/zram.sh
+    groups: [control-plane, worker]
+```
+
+You can add further entries to the list or replace the default one using the **list‑merge strategy** described earlier.
```
theboringstuff: We probably should explicitly describe the downside of our default configuration and how to disable it

docs/public/Installation.md
```diff
@@ -2583,6 +2584,71 @@ The following settings are supported in the extended format:
...
+    size: 1G
+    type: ext4
+    device: /dev/zram0
+    path: /var/log/pods
+    template:
+      source: templates/zram-setup.service.j2
+      destination: /etc/systemd/system/zram-setup.service
+    preparation_script: resources/scripts/zram.sh
+    groups: [control-plane, worker]
+```
```
theboringstuff, line 2655: Some questions:
- For an already existing environment, is it enough to just run `prepare.system.fsmount`? Will k8s pods continue to function and write logs as if nothing happened? Is reboot required?
- For ZRAM logs case, should we also highlight recommended k8s logs settings in docs, to reduce amount of RAM used? And we probably need to mention which task/procedure to run to apply k8s logs settings

kubemarine/fsmount.py
```diff
@@ -0,0 +1,201 @@
...
+
+def check_mounts(group: NodeGroup) -> List[str]:
+    """Return a list of human-readable error strings for missing or wrong-type mounts."""
+    cluster: KubernetesCluster = group.cluster
+
+    mounts_collector = CollectorCallback(cluster)
+    zramctl_collector = CollectorCallback(cluster)
+    defer = group.new_defer()
+    defer.sudo("cat /proc/mounts", callback=mounts_collector)
+    defer.sudo("zramctl --output-all", warn=True, callback=zramctl_collector)
```
theboringstuff, line 136: is it safe to run if there is no mounts at all, or no ZRAM mounts? could `zramctl` be missing in this case?

maybe we should short-circuit this check with success if there is no applicable items?

kubemarine/resources/scripts/zram.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if command -v apt-get &>/dev/null; then
+    apt-get install -yq linux-modules-extra-$(uname -r)
+elif command -v dnf &>/dev/null; then
+    dnf install -y kernel-modules-extra
+elif command -v yum &>/dev/null; then
+    yum install -y kernel-modules-extra
+fi
```
theboringstuff, line 9: I think we usually install packages from kubemarine packages functionality, it uses appropriate command, etc. Have you considered it? Or is it not applicable for this package?
- https://github.com/Netcracker/KubeMarine/blob/main/docs/public/Installation.md#management

kubemarine/templates/zram-setup.service.j2
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Create zram device
+DefaultDependencies=no
+Before=local-fs.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/sbin/modprobe zram
```
theboringstuff, line 9: we usually install modprobes in separate configuration section/task, we also have a related check, have you considered it? maybe we can add this zram module to `services.modprobe` in runtime if we see that there are `type: zram` mounts? Or is it not applicable for this module?
* https://github.com/Netcracker/KubeMarine/blob/main/docs/public/Installation.md#modprobe

docs/public/Installation.md
alexarefev: * It's not enough to run `prepare.system.fsmount`. The procedure needs to drain and reboot the node.
* I'm not sure that is the recommended way, but it really reduce the available RAM

kubemarine/templates/zram-setup.service.j2
alexarefev: The `fsmount` task is not related to the Kubernetes installation directly. We do not need `zram` kernel module on worker/control-plane node to install k8s. That is why it has been separated from usual list of kernel modules

kubemarine/fsmount.py
alexarefev: `zramctl --output-all` shows full information about all of the configured devices 
I don't understand how it could be harmful

kubemarine/resources/scripts/zram.sh
alexarefev: The `fsmount` task is not related to the Kubernetes installation directly. We do not need those packages to install k8s. Moreover, if we put that `linux-modules-extra` package into list of packages that must be installed during the `prepare` tasks it will create conjunction between different tasks. The idea is to make `fsmount` independent as much as possible to manage that part from one place of `cluster.yaml`

docs/public/Installation.md
theboringstuff: 1. So we do not have automatization for it, it should be done manually? Can you please confirm with requestor if this is ok? (I doubt, I was thinking it is expected that we provide some automatization to enable it)
2. It was requested in the ticket, so I think it is fine to use as recommendation, lets add it and describe the procedure to apply

kubemarine/fsmount.py
theboringstuff: I mean, could zramctl be missing on some fresh nodes where we do not want zram? If zramctl is always present, then it is fine to execute always

kubemarine/templates/zram-setup.service.j2
theboringstuff: ok, lets keep it here for now

kubemarine/resources/scripts/zram.sh
theboringstuff: ok, lets keep it here for now

docs/public/Installation.md
alexarefev: Where this procedure should be described? It's neither installation nor maintenance procedure

kubemarine/fsmount.py
alexarefev: Ok, I'll remove this part

docs/public/Installation.md
theboringstuff: I think it could be described in Installation procedure as an example. We already have example how to configure ZRAM, we can add note below like "note that if you use ZRAM for logs please ensure you configure these parameters ... and run this procedure ...". I think reconfigure maintenance procedure should be able to do it, but please double-check
* https://github.com/Netcracker/KubeMarine/blob/main/docs/public/Maintenance.md#reconfigure-procedure

kubemarine/fsmount.py
theboringstuff: Just to clarify, I did not mean to remove it, I mean to run it only if we actually have zram (in this case it probably safe and makes sense)

alexarefev: The `zramctl` is provided by `util-linux` package. It quite unusual when `util-linux` package is absent.

theboringstuff: ok, lets keep as is for now

docs/public/Installation.md
```diff
@@ -2583,6 +2584,71 @@ The following settings are supported in the extended format:
...
+
```
theboringstuff, line 2656: Lest please add dedicated section here for ZRAM case. Lets describe why it is used and that there are also other recommended parameters on the level of kubelet (according to ticket request).

docs/public/Maintenance.md
```diff
@@ -1357,6 +1358,72 @@ nodes:
...
+    type: zram
+    size: 1G
+    template:
+      source: templates/zram.mount.j2
+      destination: /etc/systemd/system/var-log-pods.mount
+    groups:
+      - control-plane
+      - worker
+```
+
```
theboringstuff, line 1435: For ZRAM case, lets add note here which links to installation notes ZRAM section with recommended kubelet settings and how to apply these setting on existing env

docs/public/Installation.md
alexarefev: Done

docs/public/Maintenance.md
alexarefev: Done

theboringstuff (approved):

disa1217 (approved):
@alexarefev , checked the documents.

docs/public/Maintenance.md
```diff
@@ -1357,6 +1358,87 @@ nodes:
...
+```yaml
+reboot: true
+fsmount:
+  - name: zram-pods
+    device: /dev/zram0
+    path: /var/log/pods
+    type: zram
+    size: 1G
+    template:
+      source: templates/zram.mount.j2
```
theboringstuff: Shouldn't it be `templates/zram-setup.service.j2`

PR #825, opened 2026-08-11
[CPCAP-13894] Update Envoy Gateway to 2.5.0

theboringstuff:
### Description
It is required to update Envoy Gateway version to 2.5.0 to take default BackendTrafficPolicy with timeouts configuration

### Solution
* Added Envoy Gateway 2.5.0 chart, made it the default version
* Made some changes to Envoy Gateway plugin installation
  * kubectl image is now passed to values
  * crds directoy changed for new Envoy Gateway versions
  * crds are now applied with `force-conflicts`, to avoid issues if CRDs were previously installed by another manager (e.g. ArgoCD)
  * helm is now executed with `take-ownership`, to avoid issues if KubeMarine installs chart after some other deployer
* Updated documentation

### How to apply
* run `migrate_kubemarine`

### Test Cases
1. Install cluster using previous version of KubeMarine, using Envoy Gateway as HAProxy target backend
    * make sure Envoy Gateway is of old version
2. Take new version of KubeMarine, run `migrate_kubemarine`, make sure it succeeds
    * Make sure Envoy Gateway was updated to new version
    * Run `kubectl get btp -A`, make sure there is `default-backend-traffic-policy`  in Envoy Gateway namespace
3. Install httpbin test application (from converter e2e tests)
4. Create HTTPRoute to httpbin application
    ```
    apiVersion: gateway.networking.k8s.io/v1
    kind: HTTPRoute
    metadata:
      name: test-route
      namespace: converter-test
    spec:
      hostnames:
      - httpbin.example.com
      parentRefs:
      - group: gateway.networking.k8s.io
        kind: Gateway
        name: default-external-gateway
        namespace: gateway-system
      rules:
      - backendRefs:
        - group: ""
          kind: Service
          name: httpbin
          port: 8080
          weight: 1
        matches:
        - path:
            type: PathPrefix
            value: /
    ```
6. Run following command and make sure it runs for 70 seconds uninterrupted. It means that active connection by default is not interrupted even after 15 or 60 seconds
    ```
    time curl 'https://$LB_IP:443/sse?delay=0s&duration=70s&count=70' -kv -H "Host: httpbin.example.com"
    ```
7. Run following command and make sure it get interrupted after 60 seconds. It means that IDLE connection by default is interrupted after 60 seconds
    ```
    time curl 'https://$LB_IP:443/sse?delay=70s&duration=70s&count=70' -kv -H "Host: httpbin.example.com"
    ```
8. Apply custom BackendTrafficPolicy to HTTPRoute
    ```
    apiVersion: gateway.envoyproxy.io/v1alpha1
    kind: BackendTrafficPolicy
    metadata:
      name: test-route
      namespace: converter-test
    spec:
      targetRefs:
      - group: gateway.networking.k8s.io
        kind: HTTPRoute
        name: test-route
      mergeType: StrategicMerge
      timeout:
        http:
          streamIdleTimeout: 90s
    ```
9. Run following command and make sure it runs for 100 seconds uninterrupted. It means that IDLE timeout is now more than 60s, and if connection is active it could run even more than 90s
    ```
    time curl 'https://$LB_IP:443/sse?delay=70s&duration=30s&count=30' -kv -H "Host: httpbin.example.com"
    ```

### Checklist
- [x] Integration CI passed


alexarefev (approved):

pranavcracker:
As we are going to upgrade envoy to 2.5.0, do we still need to keep the older chart (2.2.0) in the repo?

theboringstuff:
> As we are going to upgrade envoy to 2.5.0, do we still need to keep the older chart (2.2.0) in the repo?

We usually do keep older version for other plugins, so I kept it for Envoy too. Not sure if it is really needed, but for consistency I suggest to keep everything for now

pranavcracker (approved):

PR #829, opened 2026-09-09
[CPCAP-12251] ZRAM support

theboringstuff:
### Description
It is required to support ZRAM mounts to reduce disk writes, particularly for the case with `/var/log/pods`

### Solution
* Added new `services.zram` section to cluster.yaml which allows to ZRAM mount points which should be present/absent
* Task `prepare.system.modprobe` now enabled `zram` kernel module if ZRAM mount points are present
* Added new task `prepare.system.zram` which configures ZRAM mount as specified in cluster.yaml. All nodes are gracefully drained/rebooted to apply the changes. Changes are verified after reboot.
* Added new check `services.system.zram` which verifies that ZRAM configuration is consistent between cluster.yaml and nodes
* Extended `reconfigure` maintenance procedure to allow configuring required kubelet settings
* Updated documentation accordingly

### Test Cases

* Impacted procedures:
    * install
    * add_node
    * reconfiugre
    * check_paas
* Install new cluster on Ubuntu 22/24/26 with ZRAM and kubelet settings enabled, make sure ZRAM is mounted by running just `zramctl`, make sure kubelet changes applied by checking `/var/lib/kubelet/config.yaml`
* Install new cluster on Ubuntu 22/24/26 without ZRAM and kubelet settings, make sure it is not mounted using `zramctl`, make sure no kubelet changes are present.
    * Apply ZRAM/kubelet changes on this already existing cluster as described in documentation
* Remove node, rebuild it, add it back, make sure ZRAM/kubelet are configured on this node.
* Run check_paas, make sure it correctly detects discrepancies between cluster.yaml and nodes (change cluster.yaml to trigger errors)
* Try disabling ZRAM as described in documentation, make sure ZRAM is no longer mounted using `zramctl`, try enabling it back (by running just `prepare.system.zram`)

### Checklist
- [x] Integration CI passed


docs/public/Installation.md
```diff
@@ -2588,6 +2589,106 @@ The following settings are supported in the extended format:
...
+If you want to disable ZRAM on existing environment, just change ZRAM mount `state` to `absent` and run `install` procedure task `prepare.system.zram`. This task will re-configure ZRAM on all nodes and gracefully drain/reboot all nodes. Following is an example `cluster.yaml` configuration with disabled ZRAM:
+```yaml
+services:
+  zram:
+  - state: absent
+    path: /var/log/pods
+```
+
+Note that if you just remove ZRAM items from `cluster.yaml`, KubeMarine will simply stop managing them, but not remove. To properly remove ZRAM mount you need to set `state` to `absent`.
+
```
andrewluckyguy, line 2714: It would be useful to describe the procedure of changing some zram parameters, like reconfiguring the disk size

theboringstuff: added case for disk size

PR #831, opened 2026-09-18
[CPCAP-14939] configure containerd LimitNOFILE

theboringstuff:
### Description
In containerd 1.8 LimitNOFILE  soft/hard limits were changed from infinity to 1024/524288. Soft limit of 1024 is too small for some apps, so they started failing with errors like "Too many open files"

### Solution
* Added API to allow configuring containerd LimitNOFILE soft/hard limits
* Default values are 131072/524288, i.e. soft limit was increased in 128 times.
* Backup procedure was updated to also backup new containerd drop-in config.
* Kubemarine migrate patch is provided to apply on existing envs.
* Updated docs to describe new API.
* Added tests.

### How to apply
* `migrate_kubemarine`

### Test Cases

* On previous KubeMarine version, install cluster on Ubuntu OS with ingress-nginx plugin (for easy exec into pod)
* Exec into ingress-nginx pod and check soft/hard file limits, you should see 1024 (soft) and 524288 (hard) limit values
```
$ kubectl exec -it -n ingress-nginx ingress-nginx-controller-x9989 -- bash
ingress-nginx-controller-x9989:/etc/nginx$ cat /proc/1/limits | grep "Max open files"
```
* Take new KubeMarine and run `migrate_kubemarine` procedure
* Restart ingress-nginx pod
* Exec into new ingress-nginx pod and check soft/hard file limits, you should see 131072 (soft) and 524288 (hard) limit values
* Try configuring custom values for soft/hard limit (see KubeMarine updated docs on containerd), restart pod, make sure changes take effect

### Checklist
- [x] Integration CI passed


docs/public/Installation.md
```diff
@@ -2375,6 +2377,22 @@ services:
...
+You can configure this value using `cluster.yaml`, following is default configuration:
+```yaml
+services:
+  cri:
+    containerdLimitNOFILE:
+      soft: 131072
+      hard: 524288
+```
+
+Soft limit should not be less than hard, both limits can not be negative.
```
pranavcracker: I think this is inaccurate. the statement should be like - "The soft limit **should not be greater** than the hard limit" or "The soft limit **should be less** than the hard limit"

kubemarine/resources/schemas/definitions/services/cri.json
```diff
@@ -8,6 +8,9 @@
       "default": "containerd",
       "description": "A particular container runtime implementation used for Kubernetes"
     },
+    "containerdLimitNOFILE": {
+      "$ref": "#/definitions/ContainerdConfig"
```
pranavcracker: Is the value for **$ref** correct?
I think it should be - `containerdLimitNOFILE`

pranavcracker (changes requested):

docs/public/Installation.md
theboringstuff: fixed

kubemarine/resources/schemas/definitions/services/cri.json
theboringstuff: fixed

pranavcracker (approved):

disa1217 (approved):
@pranavcracker , reviewed.

